PR #13, opened 2016-05-22
Resource Filtering

jaime29010:


DemonWav:
I'm fine with the first commit, but I'm not quite sure about the second. The wording feels more awkward to me, but I'll see if I can get more input on that from others.

DemonWav:
You need to enable filtering with Gradle as well, not just Maven.

jaime29010:
Okay, I'll add filtering with Gradle in the next commit.

jaime29010:
I have tested it and it works, it replaces ${VERSION} with @version@ or ${project.version} respectively on the BuildSystem of the module

DemonWav:
You already have the BuildSystem in the #create() method of ProjectConfiguration, no need to get a new instance in the template util class.

jaime29010:
Couldn't the "When To Load This Plugin" be changed to something else? Like "Loading stage" or something else. Or just "When to load this plugin".

PR #27, opened 2016-07-08
Use shields.io for the TeamCity badge

JackDn:


DemonWav:
I'm not against this change, I'm just a little curious as to why or what the benefit is?

kashike:
It looks a bit better in my opinion, matches the others better.

DemonWav:
Good enough for me.

PR #30, opened 2016-07-11
Changed values in @Mod annotation of forge template to rely on variables

Redrield:
It is customary for Forge developers to define the Mod ID, the name, and the version of their mod in constants defined in the main class. This is so that they can change all references to them globally, and they don't need to constantly look for the old values when, say, updating the version of the mod.

### DIFF
--- a/src/main/java/com/demonwav/mcdev/buildsystem/gradle/GradleBuildSystem.java
+++ b/src/main/java/com/demonwav/mcdev/buildsystem/gradle/GradleBuildSystem.java
@@ -45,7 +45,11 @@ import com.intellij.openapi.util.Pair;
 import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VfsUtil;
 import com.intellij.openapi.vfs.VirtualFile;
-import com.intellij.psi.*;
+import com.intellij.psi.PsiDirectory;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiFile;
+import com.intellij.psi.PsiFileFactory;
+import com.intellij.psi.PsiManager;
 import org.apache.commons.io.FileUtils;
 import org.gradle.tooling.BuildLauncher;
 import org.gradle.tooling.GradleConnector;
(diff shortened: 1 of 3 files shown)


src/main/java/com/demonwav/mcdev/buildsystem/gradle/GradleBuildSystem.java
```diff
@@ -53,11 +45,7 @@
 import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VfsUtil;
 import com.intellij.openapi.vfs.VirtualFile;
-import com.intellij.psi.PsiDirectory;
-import com.intellij.psi.PsiElement;
-import com.intellij.psi.PsiFile;
-import com.intellij.psi.PsiFileFactory;
-import com.intellij.psi.PsiManager;
+import com.intellij.psi.*;
```
gabizou: Don't star import.

src/main/java/com/demonwav/mcdev/buildsystem/gradle/GradleBuildSystem.java
```diff
@@ -88,15 +73,7 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
gabizou: As above.

src/main/java/com/demonwav/mcdev/creator/ForgeProjectSettingsWizard.java
```diff
@@ -7,7 +7,6 @@
 import com.demonwav.mcdev.platform.forge.versionapi.McpVersion;
 import com.demonwav.mcdev.platform.forge.versionapi.McpVersionEntry;
 import com.demonwav.mcdev.platform.hybrid.SpongeForgeProjectConfiguration;
-
```
gabizou: Don't change the format for this.

PR #33, opened 2016-07-22
Adding mappings of TextFormatting => Color for Forge module

Redrield:
I added the gutter color mappings for the TextFormatting enum for Minecraft Forge mods


src/main/java/com/demonwav/mcdev/platform/forge/ForgeModuleType.java
```diff
@@ -69,4 +72,29 @@ public String getId() {
     public ForgeModule generateModule(Module module) {
         return new ForgeModule(module);
     }
+
+    private void setupMap() {
```
kashike: This can just go in the constructor.

Issue #37, opened 2016-07-25
Forge Run Configs should use GradleStart


TehNut:
Instead of launching using `runClient` and `runServer`, Forge mods should launch as an Application using `GradleStart` and `GradleStartServer`. 

* This should use the classpath of the `PROJECT_main` module.
* The working directory should default to `PROJECT_DIR\run`.

![Example](http://i.imgur.com/MhIR4Rl.png)

DemonWav:
I agree, this was part of a PR that did something else and I just accepted it as is. This will be fixed at some point, hopefully soon.

PR #40, opened 2016-07-31
Fixed the problem with run Client/Server configurations

Redrield:
Fixed the issue addressed by [#37](https://github.com/DemonWav/MinecraftDevIntelliJ/issues/37)

### DIFF
--- a/src/main/java/com/demonwav/mcdev/buildsystem/gradle/GradleBuildSystem.java
+++ b/src/main/java/com/demonwav/mcdev/buildsystem/gradle/GradleBuildSystem.java
@@ -316,9 +316,10 @@ public class GradleBuildSystem extends BuildSystem {
                     if (!runningDir.exists()) {
                         runningDir.mkdir();
                     }
+                    Module mainModule = ModuleManager.getInstance(project).findModuleByName(module.getName() + "_main");
                     runClientConfiguration.setWorkingDirectory(project.getBasePath() + File.separator + "run");
                     runClientConfiguration.setMainClassName("GradleStart");
-                    runClientConfiguration.setModule(ModuleManager.getInstance(project).findModuleByName(module.getName() + "_main") != null ? ModuleManager.getInstance(project).findModuleByName(module.getName() + "_main") : module);
+                    runClientConfiguration.setModule(mainModule != null ? mainModule : module);
                     RunnerAndConfigurationSettings clientSettings = new RunnerAndConfigurationSettingsImpl(
                             RunManagerImpl.getInstanceImpl(project),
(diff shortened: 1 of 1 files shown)


src/main/java/com/demonwav/mcdev/buildsystem/gradle/GradleBuildSystem.java
```diff
@@ -305,15 +307,18 @@ public void finishSetup(@NotNull Module module, @NotNull ProjectConfiguration co
...
-                    runClientConfiguration.getSettings().setExternalProjectPath(rootDirectory.getPath());
-                    runClientConfiguration.getSettings().setExecutionName(module.getName() + " run client");
-                    runClientConfiguration.getSettings().setTaskNames(Collections.singletonList("runClient"));
+                    File runningDir = new File(project.getBasePath(), "run");
+                    if (!runningDir.exists()) {
+                        runningDir.mkdir();
+                    }
+                    runClientConfiguration.setWorkingDirectory(project.getBasePath() + File.separator + "run");
+                    runClientConfiguration.setMainClassName("GradleStart");
+                    runClientConfiguration.setModule(ModuleManager.getInstance(project).findModuleByName(module.getName() + "_main"));
```
DemonWav: This is only true if the user chose to use separate modules for source sets. You need to handle the case where there is only one module.

src/main/java/com/demonwav/mcdev/buildsystem/gradle/GradleBuildSystem.java
```diff
@@ -305,15 +307,18 @@ public void finishSetup(@NotNull Module module, @NotNull ProjectConfiguration co
...
-                    runClientConfiguration.getSettings().setExternalProjectPath(rootDirectory.getPath());
-                    runClientConfiguration.getSettings().setExecutionName(module.getName() + " run client");
-                    runClientConfiguration.getSettings().setTaskNames(Collections.singletonList("runClient"));
+                    File runningDir = new File(project.getBasePath(), "run");
+                    if (!runningDir.exists()) {
+                        runningDir.mkdir();
+                    }
+                    runClientConfiguration.setWorkingDirectory(project.getBasePath() + File.separator + "run");
+                    runClientConfiguration.setMainClassName("GradleStart");
+                    runClientConfiguration.setModule(ModuleManager.getInstance(project).findModuleByName(module.getName() + "_main") != null ? ModuleManager.getInstance(project).findModuleByName(module.getName() + "_main") : module);
```
DemonWav: Don't do the module lookup twice. Also, only go past the 140 column line limit if breaking would hurt readability.

PR #64, opened 2016-09-24
Use SpongeGradle for Sponge plugin Gradle projects

Minecrell:
[SpongeGradle](https://docs.spongepowered.org/master/en/plugin/project/gradle.html#using-spongegradle) is a Gradle plugin that automatically sets a lot of defaults for Sponge plugins using Gradle. The primary feature is that it automatically sets the plugin version defined in the Gradle build file to the generated plugin metadata, so plugin authors will only need to update their version in one place.

This PR changes the Sponge Gradle templates to automatically apply SpongeGradle and removes all settings that are automatically applied by SpongeGradle.

Let me know if any changes are necessary before this can be merged. :)


DemonWav:
Looks good, thanks.

PR #74, opened 2016-10-08
Add inspection for references to Mixin classes

Aaron1011:


src/main/java/com/demonwav/mcdev/platform/mixin/inspections/MixinClassReferenceInspection.java
```diff
@@ -0,0 +1,64 @@
+package com.demonwav.mcdev.platform.mixin.inspections;
```
kashike: Missing licence leader.

JBYoshi:
Would this skip cases such as one mixin extending another where the reference gets replaced at runtime?

DemonWav:
This isn't mergeable as it is, @JBYoshi, I've told @Aaron1011 what needs to be changed for it to work right. Currently it doesn't check for extends or implements, but it will when it's pulled.

DemonWav:
Merged.

PR #103, opened 2016-12-03
Use the class selected with the cursor for the Find Mixins action

Minecrell:
Assuming I have a Mixin class open in the editor:

```java
@Mixin(Entity.class)
public abstract class MixinEntity {

}
```

When I right click the `Entity.class` part I'd expect the `Find mixins` action to search for other Mixins that target `Entity` and not for Mixins which target `MixinEntity`. With these changes it will attempt to resolve the selected reference before checking for mixins of the current class.

Since the changed method is also used in other contexts I'm not entirely sure if the change also breaks something else.

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/util/McPsiUtil.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/util/McPsiUtil.kt
@@ -26,33 +26,31 @@ import com.intellij.psi.PsiFile
 import com.intellij.psi.PsiMember
 import com.intellij.psi.PsiModifier
 import com.intellij.psi.PsiModifierListOwner
+import com.intellij.psi.PsiReference
 import com.intellij.psi.search.GlobalSearchScope
 import org.jetbrains.annotations.Contract
 import java.util.Collections
 import java.util.stream.Collectors
 
 @Contract(value = "null -> null", pure = true)
-fun getClassOfElement(element: PsiElement?): PsiClass? {
-    if (element == null) {
-        return null
-    }
-
-    if (element is PsiClass) {
-        return element
-    }
-
+fun getClassOfElement(element: PsiElement?, resolveReferences: Boolean = false): PsiClass? {
     var el = element
-    while (el?.parent != null) {
-        if (el.parent is PsiClass) {
-            return el.parent as PsiClass
+    while (el != null) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/demonwav/mcdev/util/McPsiUtil.kt
```diff
@@ -17,42 +17,31 @@ import com.intellij.navigation.AnonymousElementProvider
...
-import com.intellij.psi.JavaPsiFacade
-import com.intellij.psi.PsiAnnotation
-import com.intellij.psi.PsiClass
-import com.intellij.psi.PsiDirectory
-import com.intellij.psi.PsiElement
-import com.intellij.psi.PsiFile
-import com.intellij.psi.PsiMember
-import com.intellij.psi.PsiModifier
-import com.intellij.psi.PsiModifierListOwner
+import com.intellij.psi.*
```
DemonWav: No * imports.

src/main/kotlin/com/demonwav/mcdev/util/McPsiUtil.kt
```diff
@@ -17,42 +17,31 @@ import com.intellij.navigation.AnonymousElementProvider
...
 import com.intellij.psi.search.GlobalSearchScope
 import org.jetbrains.annotations.Contract
-import java.util.Collections
+import java.util.*
```
DemonWav: No * imports.

Minecrell:
@DemonWav Fixed

DemonWav:
I'm not comfortable with this implementation, as getClassOfElement()'s purpose is specifically to get the class that _contains_ the element, never the class _of_ the element. I think this should be implemented in [FindMixinsAction](https://github.com/minecraft-dev/MinecraftDev/blob/2016.3/src/main/java/com/demonwav/mcdev/platform/mixin/actions/FindMixinsAction.java) instead.

Minecrell:
@DemonWav I've modified `getClassOfElement` to take an additional `resolveReferences` parameter which is `false` by default. Does that solve the issue?

DemonWav:
[Manually fixed and pulled.](https://github.com/minecraft-dev/MinecraftDev/commit/656be500e574d126b8b2e8936aea5116e0960295)

PR #107, opened 2016-12-07
Use GradleProjectResolverExtension to obtain data for the SRG mappings

Minecrell:
`GradleProjectResolverExtension` allows hooking the Gradle refresh action in the IDE to inject additional models used by IntelliJ IDEA plugins. We can use it to completely remove the need for an additional Gradle action every time the project is opened and sync our data together with the dependency resolution of IntelliJ.

That should reduce the overall initialization time and has the advantage that it uses the Gradle options made in IntelliJ, e.g. using a local distribution instead of the Gradle wrapper.

~~This is primarily WIP because it needs some cleanup. Additionally we would need to store the resolved data together with the dependencies so you don't need to refresh the Gradle project every time you open a project.~~


Minecrell:
@DemonWav 
I've replaced most of the classes with Kotlin and store the resolved data in the module configuration now. It works fine as far as I've tested. Please let me know what you think! ;)

src/main/kotlin/com/demonwav/mcdev/platform/mcp/gradle/McpModelData.kt
```diff
@@ -0,0 +1,44 @@
...
+
+class McpModelData(val module: ModuleData, val settings: McpModuleSettings.State) : AbstractExternalEntityData(module.owner) {
+
+    companion object {
+        // Process MCPModelData after builtin services (e.g. dependency or module data)
+        val KEY = Key.create(McpModelData::class.java, ProjectKeys.TASK.processingWeight + 1)
+    }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
```
DemonWav: https://www.demonwav.com/style#s4.1.1-braces-always-used

src/main/kotlin/com/demonwav/mcdev/platform/mcp/srg/SrgManager.kt
```diff
@@ -0,0 +1,42 @@
...
+    var srgMap: Promise<SrgMap> = rejectedPromise("SRG map not loaded")
+        @Synchronized get
+        private set
+
+    val srgMapNow: SrgMap?
+        @Synchronized get() = (srgMap as? Getter<*>)?.get() as SrgMap?
+
+    @Synchronized
+    fun parse(files: Set<String>) {
+        srgMap = if (files.isNotEmpty())
```
DemonWav: https://www.demonwav.com/style#s4.1.1-braces-always-used

DemonWav:
YOLO

PR #108, opened 2016-12-08
Convert FindMixinsAction to Kotlin and use AnnotatedElementsSearch

Minecrell:
`AnnotatedElementsSearch` allows searching for elements with a specific annotation directly so we don't need to check all classes in the project for the Mixin annotation. While I don't have any proof it seemed to improve performance a bit but the difference is probably minor.

Converted the action to Kotlin to make the code nicer.


gabizou:
@Minecrell one thing that is for sure, I did have in mind to cache these results so that we can start caching other results, such as injectors or overwrites etc.

PR #110, opened 2016-12-11
Removed line that registers listeners...

Redrield:
...on main class, as it causes errors in the IDE because the class does not implement Listener


src/main/resources/fileTemplates/j2ee/bukkit_main_class.java.ft
```diff
@@ -7,7 +7,7 @@ public final class ${CLASS_NAME} extends JavaPlugin {
     @Override
     public void onEnable() {
         // Plugin startup logic
-        this.getServer().getPluginManager().registerEvents(this, this);
+
```
jamierocks: Haven't even removed the line. An alternative would be to implement Listener.

PR #114, opened 2016-12-17
Add copy mixin target reference action

Minecrell:
Adds a new action that allows you to right click a reference to a field/method while browsing the Minecraft source and copies the target reference you would use in an `@Inject` to the clipboard.

Supported elements:
- Fields, e.g. `Lnet/minecraft/block/Block;translucent:Z`
- Methods, e.g. `Lnet/minecraft/block/Block;disableStats()Lnet/minecraft/block/Block;`

Not sure about the name for the action, feel free to suggest a better one.


loordgek:
why closed ?

Minecrell:
@loordgek Moved to #118

PR #115, opened 2016-12-17
Add Mixin debugging support

Minecrell:
Adds support for debugging Mixins the same way you can debug regular Java classes. See https://github.com/SpongePowered/Mixin/pull/155 for an explanation of how it works.

You will still not see the invocation of the injector while debugging, however you can step in the Mixin injector and/or set a breakpoint in a Mixin method.

### Implementation
There are 3 parts that implement the Mixin debugging support for IntelliJ IDEA:

- `MixinRunConfigurationExtension`: Automatically adds the Mixin option (`-Dmixin.sourceDebugExtension=true`) to the VM args when debugging a Mixin module.
- `MixinPositionManager`: Parses the `SourceDebugExtension` added by Mixin and returns the correct Mixin source file while debugging. It will also redirect all breakpoints to the target class so they will work at runtime.
- `MixinDebuggerClassFilterProvider`: Excludes `CallbackInfo` classes from the debugger. Since you don't see the invocation of the injector you need to use Step Into where you expect it to be, the constructor of `CallbackInfo` was just an additional confusing source it jumps to before going to the injector.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -75,6 +75,9 @@ dependencies {
         exclude(group = "org.jetbrains", module = "annotations")
     }
 
+    // Add tools.jar for the JDI API
+    compile(files(org.gradle.internal.jvm.Jvm.current().toolsJar))
+
     // Add an additional dependency on kotlin-runtime. It is essentially useless
     // (since kotlin-runtime is a transitive dependency of kotlin-stdlib-jre8)
     // but without kotlin-stdlib or kotlin-runtime on the classpath,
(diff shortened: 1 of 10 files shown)


DemonWav:
0/10 short branch name.

kashike:
https://github.com/SpongePowered/Mixin/issues/156

gabizou:
> 0/10 short branch name.

Clearly you haven't seen the branch names on Sponge.

Minecrell:
Changes in Mixin have been pushed, this is ready for merge now.

Issue #102, opened 2016-12-03
Mixin classes can be referenced if they are part of the hierarchy


Minecrell:
Currently, MinecraftDev reports an error every time Mixin classes are referenced directly. That is usually correct, because they do not exist at runtime. However, it is possible to reference them in Mixins if they are part of its class hierarchy.

MinecraftDev shouldn't mark the usages as errors if they are part of the Mixin's hierarchy.

Real example that causes an error with MinecraftDev: https://github.com/SpongePowered/SpongeVanilla/blob/bleeding/src/main/java/org/spongepowered/server/mixin/core/entity/player/MixinEntityPlayer.java#L143-L145

Another example that shows how it works:

```java
@Mixin(Block)
public class MixinBlock {
     public int blockInt;
}

@Mixin(Entity.class)
public class MixinEntity {
    public int veryCoolField = 1;
}

@Mixin(EntityPlayer.class)
public class MixinEntityPlayer extends MixinEntity {

    public void veryCoolMethod() {
        System.out.println(this.veryCoolField);
    }

    private void doSomething(EntityPlayer player, Block block) {
        // We can cast to MixinEntity to access the field
        System.out.println("About to print " + ((MixinEntity) (Object) player).veryCoolField);

        // ... or we can also cast to MixinEntityPlayer to call the method
        ((MixinEntityPlayer) (Object) player).veryCoolMethod();

        // ERROR: We can't reference MixinBlock because it is not part of the hierarchy
        int i = ((MixinBlock) (Object) block).blockInt;
    }
}
```

Minecrell:
Fixed in #118 (https://github.com/minecraft-dev/MinecraftDev/pull/118/commits/f393097961833b82b49384d8a90cf074e125db98)

PR #118, opened 2016-12-18
Improved Mixin code assistance

Minecrell:
These changes extend MinecraftDev with a lot of utilities for working with Mixins:

- **Automatic completion of `@Shadow` members:** Members in the target class are available like regular fields in the code completion of the Mixin class. If selected, a `@Shadow` member will be generated automatically.
- Code completion + inspection for method reference in `@Inject`, `@ModifyArg`, `@ModifyConstant`, `@ModifyVariable` and `@Redirect`, as well as injection point types
- "Generate `@Overwrite`" action: Copies the original method source and adds all necessary `@Shadow` members
- "Generate `@Shadow`" action: Generates a `@Shadow` member for a selected member in the target class
- `@Overwrite` inspection and line markers (to verify that the target method actually exists)
- Improved handling of copy/paste from the target class, will automatically add all necessary `@Shadow` elements when copying code from the target class
- Inspection for `@Inject` method signatures ("should be (not) static" and method parameters)
- Fix for #102 - allow references to Mixins if they are part of the hierarchy
- Static member inspection now ignores `@Shadow` members
- "Copy Mixin target reference" action that copies the qualified descriptor for the `target` attribute of injection annotations
- **WIP:** Code completion for `target` attribute of injection annotations
- Possibly other changes I can't remember.

**Planned changes (maybe):**
- Report unnecessary qualified descriptors
- Inspection for method signature of other injection point types + improved quick fixes
- Unit tests


DemonWav:
Did this really have to be in one giant PR? :S

Minecrell:
@DemonWav Yes, most of this uses shared code, it would be difficult to separate.

gabizou:
@Minecrell the issue with this giant pr is that while it does add a lot of features, it's near impossible to review with limited time. In @DemonWav's defense, his time is fairly limited as is.

Minecrell:
Moved to #123, #124 and #125

PR #123, opened 2016-12-31
Add base classes for improved Mixin code assistance

### LINKED ISSUE #102: Mixin classes can be referenced if they are part of the hierarchy

Minecrell:
These changes add the base classes/methods that are used in #124 and #125.

I've also ported `MixinClassReferenceInspection` to Kotlin and fixed #102.


src/main/kotlin/com/demonwav/mcdev/util/McPsiUtil.kt
```diff
@@ -35,26 +35,70 @@ import java.util.stream.Collectors
...
+
+    return null
+}
+
+inline fun findLastChild(element: PsiElement, condition: (PsiElement) -> Boolean): PsiElement? {
+    var child = element.firstChild
+    var lastChild: PsiElement? = null
+
+    while (child != null) {
+        if (condition.invoke(child)) {
```
DemonWav: I don't know if this is really better stylistically, but you can actually treat this just as function (`condition(child)` rather than `condition.invoke(child)`) and I personally like treating it as a function better, as it seems a little more expressive as to what it actually is (or more accurately, what it is intended to do).

src/main/kotlin/com/demonwav/mcdev/util/Util.kt
```diff
@@ -35,6 +35,32 @@ fun invokeLater(func: () -> Unit) {
     ApplicationManager.getApplication().invokeLater { func() }
 }
 
+/**
+ * Returns an untyped array for the specified [Collection].
+ */
+internal fun Collection<*>.toArray(): Array<Any> {
```
DemonWav: This isn't used in this PR, but I'm assuming it's used in the other two PR's.

I think this is dangerous, though. You're creating an `Array<Any>`, so this array cannot hold nullable values. `java.util.Collection#toArray()` returns (from Kotlin's perspective) `Array<Any!>!` which means it will allow treating it as either nullable or null-safe, since it comes from un-annotated Java code, so there is no way for it to know either way. If we know the contract on the method is null-safe, we can happily treat it as such, but that isn't the case here. That means if this is used and `#toArray()` returns an array with nulls, this may cause failures in Kotlin's nullability checks later down the line, since it's assumed to be non-null only.

Minecrell: You're right, it should return `Array<Any?>`

src/main/kotlin/com/demonwav/mcdev/util/McPsiUtil.kt
Minecrell: I agree, I didn't think about that when writing this code

src/main/kotlin/com/demonwav/mcdev/util/McBytecodeUtil.kt
```diff
@@ -0,0 +1,112 @@
...
+        PsiType.SHORT -> 'S'
+        PsiType.BOOLEAN -> 'Z'
+        PsiType.VOID -> 'V'
+        else -> throw IllegalArgumentException("Unsupported primitive type: $this")
+    }
+
+private fun PsiClassType.erasure() = TypeConversionUtil.erasure(this) as PsiClassType
+
+internal val PsiClassType.internalName
+    get() = erasure().resolve()!!.internalName
```
Minecrell: I see your point. The reason I've made these properties throw an exception is mostly to simplify the code and to keep it consistent. There are a lot of other methods where this value is required to do certain checks or to build a part of the result (e.g. the descriptor generation for a specific method/field). For some of them I can't simply return null. 

For example, directly below this property we have the `appendInternalName` method which takes a `StringBuilder`, appends the internal name and returns the builder again. I cannot return null for this method, so we would have one method which returns null and one method which throws an exception.

It gets more complicated once we get to the descriptors. It also has a property and an append method, but at the end the implementation of both methods uses the `appendInternalName` method - how would we return null for the descriptor if one of the classes cannot be resolved?

**TL;DR:** The reason why I didn't make it return null is that, at the end, all code that is using these properties/methods requires the descriptors to continue successfully. In my testing I've never encountered a case where `resolve()` returned null. These cases would need special handling instead of just failing silently so throwing an exception was the better solution to allow us to handle these cases more leniently in the future.

src/main/kotlin/com/demonwav/mcdev/util/McPsiClass.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.intellij.navigation.AnonymousElementProvider
+import com.intellij.openapi.extensions.Extensions
+import com.intellij.psi.PsiClass
+import com.intellij.psi.PsiClassType
+import com.intellij.psi.PsiElement
+
+// Type
+
+internal val PsiClassType.fullQualifiedName
+    get() = resolve()!!.fullQualifiedName
```
Minecrell: I want to add that some of the assert non-null calls here are actually correct. For example, [if a `PsiClass` doesn't have a `containingClass` it cannot be an anonymous class, therefore it should always have a `qualifiedName`](https://github.com/minecraft-dev/MinecraftDev/pull/123/files#diff-2f51dab0b9d851253f3671e2a778ecc0R29).

src/main/kotlin/com/demonwav/mcdev/util/McBytecodeUtil.kt
DemonWav: The issue with your assumption is uncaught exceptions in IntelliJ don't fail silently. A bug notification pops up, with the stacktrace, and information on where the exception originated (this plugin).

> For example, directly below this property we have the appendInternalName method which takes a StringBuilder, appends the internal name and returns the builder again. I cannot return null for this method, so we would have one method which returns null and one method which throws an exception.

If it returns null then just don't append to the builder and return it? Either that, or throw a _specific_ exception so the actual failure can be handled, don't just throw a NPE (or cause an NPE to be thrown).

> how would we return null for the descriptor if one of the classes cannot be resolved?

In this case, like the above, a return value of `null` is either the fail case (which would then allow the _caller_ to handle the issue, not the utility method), or a specific exception should be thrown, rather than an ambiguous NPE. Either way, it should be in the contract of the method, and the caller should deal with it, but the caller should _never_ be catching a NPE because the utility method doesn't provide any other way to handle that error condition.

Minecrell: > The issue with your assumption is uncaught exceptions in IntelliJ don't fail silently. A bug notification pops up, with the stacktrace, and information on where the exception originated (this plugin).

That is correct, and once we have such an exception we can specifically modify the code to handle the situations in which the `resolve` method can return null.

> If it returns null then just don't append to the builder and return it? Either that, or throw a specific exception so the actual failure can be handled, don't just throw a NPE (or cause an NPE to be thrown).

Returning without appending anything would be a bad idea because you would run into many seemingly unrelated errors in the code because the result of these methods is often used for comparison of two method signatures. I agree that a NPE is not a good solution because of its ambiguity but I believe it is a good temporary solution until we can properly handle these situations.

This is essentially one case were I had to keep the code simple by making certain assumptions (e.g. here: that it will always `resolve` correctly) because other features took much longer than expected because of several special cases I had to handle. This is one of the parts I'd like to improve but I won't have much time for this in the next weeks.

DemonWav: I'll see if I can fix it up, then, before I pull. In the instances where these are used, `resolve` shouldn't ever return null, but I promise you it can and will. I understand your reasoning for why you did it this way, but I can't pull this PR like that. I don't think it will be very difficult to handle these situations (or at the very least just continuing when the errors are thrown for now) so I don't think I'll have too much of an issue fixing these up.

src/main/kotlin/com/demonwav/mcdev/util/Util.kt
Minecrell: Fixed

src/main/kotlin/com/demonwav/mcdev/util/McPsiUtil.kt
Minecrell: Fixed

PR #125, opened 2016-12-31
Add Mixin reference completion

Minecrell:
Adds code completion and inspections for references in the Mixin injector annotations.

**New features:**
- Code completion + inspection for method reference in `@Inject`, `@ModifyArg`, `@ModifyConstant`, `@ModifyVariable` and `@Redirect`
    - Inspection to report ambiguous references
    - Inspection to report unnecessary qualified references
- Code completion + inspection for target references of the `@At` annotation
    - Supports `INVOKE`, `INVOKE_ASSIGN`, `INVOKE_STRING`, `FIELD` and `NEW`
    - Inspection for unqualified references
- Code completion + inspection for injection point types (`INVOKE`, `FIELD`, etc)
- Inspection for `@Inject` and `@Redirect` method signatures ("should be static", method parameters, return type)
- "Copy Mixin target reference" action that copies the qualified descriptor for the `target` attribute of the `@At` annotation

Short video showing code completion (click image to open it):

[](https://download.minecrell.net/minecraft/sponge/test.mp4)

**Note:** This PR targets the `mixin-base` branch so you only see the new changes instead of also all the base classes. #123 is required for this PR.

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mixin/inspection/reference/InvalidMemberReferenceInspection.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mixin/inspection/reference/InvalidMemberReferenceInspection.kt
@@ -1,0 +1,62 @@
+/*
+ * Minecraft Dev for IntelliJ
+ *
+ * https://minecraftdev.org
+ *
+ * Copyright (c) 2017 minecraft-dev
+ *
+ * MIT License
+ */
+
+package com.demonwav.mcdev.platform.mixin.inspection.reference
+
+import com.demonwav.mcdev.platform.mixin.inspection.MixinInspection
+import com.demonwav.mcdev.platform.mixin.reference.MethodReference
+import com.demonwav.mcdev.platform.mixin.reference.MixinReference
+import com.demonwav.mcdev.platform.mixin.reference.target.TargetReference
+import com.demonwav.mcdev.platform.mixin.util.MemberReference
+import com.demonwav.mcdev.util.annotationFromNameValuePair
+import com.demonwav.mcdev.util.constantStringValue
+import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.psi.JavaElementVisitor
+import com.intellij.psi.PsiElementVisitor
+import com.intellij.psi.PsiNameValuePair
+
+class InvalidMemberReferenceInspection : MixinInspection() {
+
+    override fun getStaticDescription() = """
+        |Reports invalid usages of member references in Mixin annotations. Two different formats are supported by Mixin:
+        | - Lcom/example/ExampleClass;execute(II)V
+        | - com.example.ExampleClass.execute(II)V
+    """.trimMargin()
+
+    override fun buildVisitor(holder: ProblemsHolder): PsiElementVisitor = Visitor(holder)
+
+    private class Visitor(private val holder: ProblemsHolder) : JavaElementVisitor() {
+
+        override fun visitNameValuePair(pair: PsiNameValuePair) {
+            val name = pair.name ?: return
+
+            val resolver: MixinReference = when (name) {
+                "method" -> MethodReference
+                "target" -> TargetReference
+                else -> return
+            }
+
+            // Check if valid annotation
+            val annotation = pair.annotationFromNameValuePair ?: return
+            if (!resolver.isValidAnnotation(annotation.qualifiedName!!)) {
+                return
+            }
+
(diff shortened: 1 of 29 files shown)


kashike:
There are a bunch of `todo`s still in the code - are they intentionally not done yet, or is this still a work in progress?

src/main/kotlin/com/demonwav/mcdev/platform/mixin/inspection/MixinClassReferenceInspection.kt
```diff
@@ -41,7 +41,7 @@ class MixinClassReferenceInspection : MixinInspection() {
                 return
             }
 
-            holder.registerProblem(type, "${referencedClass.name} is a Mixin class, and cannot be referenced directly")
+            holder.registerProblem(type, "Mixin class cannot be referenced directly")
```
kashike: Why?

Minecrell:
@kashike
They are intentionally not done. I've decided to focus on the primary features first because I don't have enough time currently to implement some of the more complex features like full support for Mixins with multiple targets. Basically these are open for contributions by others or myself (but at a later point). I can't work on them in the next weeks.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/inspection/MixinClassReferenceInspection.kt
Minecrell: This should be really in #123, not sure why this is here. Anyway, the reason I did this is because the message seemed a bit long to me, in most cases you have the class name directly highlighted in the editor anyway (no need to display it again).

Mumfrey: Not sure if this is already handled elsewhere, but certain mixin classes *can* be referenced directly. Specifically *Accessor Mixins*.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/inspection/reference/InvalidMemberReferenceInspection.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.intellij.psi.PsiElementVisitor
+import com.intellij.psi.PsiNameValuePair
+
+class InvalidMemberReferenceInspection : MixinInspection() {
+
+    override fun getStaticDescription() = """
+        |Reports invalid usages of member references in Mixin annotations. Two different formats are supported by Mixin:
+        | - Lcom/example/ExampleClass;execute(II)V
+        | - com.example.ExampleClass.execute(II)V
+    """
```
DemonWav: This is missing the `.trimMargin()` call.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/inspection/MixinClassReferenceInspection.kt
Minecrell: I didn't add this inspection (I've just ported it to Kotlin), but one case that was broken was if referenced Mixin classes were either the Mixin itself or one of its super mixins (https://github.com/minecraft-dev/MinecraftDev/issues/102). I've only fixed that case currently. I don't know much about accessor Mixins so this is probably not handled currently.

DemonWav: I don't believe Aaron had accessor Mixins in mind when he wrote that inspection.

Mumfrey: For your reference then:

*Accessor Mixins* are the only case currently where it's legal to refer to a Mixin class *outside* of the mixin hierarchy and they can be identified by:

* Mixin itself must be an `interface`
* All mixin targets must be `class`es
* All member methods must be decorated with `@Accessor` or `@Invoker`

If the above conditions are met then the Mixin is applied as an *Accessor Mixin* and is allowed to be referenced from consumer code:

```java
@Mixin(Entity.class)
public interface MixinEntity {

    @Accessor
    public abstract Set<String> getTags();

}
```

In consumer code:

```java
Entity entity = this.getSomeEntity();
// Note that this cast is legal because the mixin is an accessor
Set<String> entityTags = ((MixinEntity)entity).getTags();
```

Minecrell:
🎉

PR #157, opened 2017-02-09
Ignore case when looking up gradle modules for sponge

ewized:
This has fixed a bug for myself where my project has the gradle dependency links in lowercase.


DemonWav:
f9cf758

PR #169, opened 2017-03-03
Speed up @SidedProxy indexing.

JBYoshi:
This replaces the "all classes" search for `@SidedProxy` fields with the built-in `AnnotatedElementsSearch`, which is a lot faster.

It also makes the progress bar show actual progress. It's just in case someone has a lot of `@SidedProxy` fields. :wink:

### DIFF
--- a/src/main/java/com/demonwav/mcdev/platform/forge/inspections/sideonly/SideOnlyProjectComponent.java
+++ b/src/main/java/com/demonwav/mcdev/platform/forge/inspections/sideonly/SideOnlyProjectComponent.java
@@ -60,7 +60,7 @@ public class SideOnlyProjectComponent extends AbstractProjectComponent {
                             if (sidedProxy == null) {
                                 return;
                             }
-                            Collection<PsiField> annotatedFields = AnnotatedElementsSearch.searchElements(sidedProxy, scope, PsiField.class).findAll();
+                            Collection<PsiField> annotatedFields = AnnotatedElementsSearch.searchPsiFields(sidedProxy, scope).findAll();
                             indicator.setIndeterminate(false);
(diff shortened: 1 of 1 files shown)


DemonWav:
Thanks buddy, I plan on fundamentally redoing how this is actually handled at some point, but this is an excellent stopgap between now and then.

src/main/java/com/demonwav/mcdev/platform/forge/inspections/sideonly/SideOnlyProjectComponent.java
```diff
@@ -53,25 +55,19 @@ public void projectOpened() {
...
-                                    }
-                                );
-                            } catch (Exception ignored2) {}// I have no idea, seems like it's a bug with intellij, but I'll catch any
-                                                           // weird exceptions that pop up anyways...
+                            GlobalSearchScope scope = GlobalSearchScope.projectScope(myProject);
+                            PsiClass sidedProxy = JavaPsiFacade.getInstance(myProject).findClass(ForgeConstants.SIDED_PROXY_ANNOTATION, scope);
+                            if (sidedProxy == null) {
+                                return;
+                            }
+                            Collection<PsiField> annotatedFields = AnnotatedElementsSearch.searchElements(sidedProxy, scope, PsiField.class).findAll();
```
Minecrell: You can use `AnnotatedElementsSearch.searchPsiFields` here instead of adding `PsiField.class` explicitely

JBYoshi:
@Minecrell Fixed.

...Never mind. I think you beat me to the web page. :laughing:

JBYoshi:
@DemonWav Fixed.

Issue #163, opened 2017-02-14
[Forge] Add an Alt-Enter shortcut to add a private/protected field/method to the AT.


me4502:
Being able to automatically add to AT from the Alt+Enter menu could be potentially useful.

DemonWav:
The intention is to add class and method/field auto-completion, as well as maybe adding an entry automatically from the member like what you're suggesting.

Issue #49, opened 2016-08-21
Add Access Transformer shortcuts


ZephireNZ:
My "wishlist":
- Clicking through a Access Transformer to it's location in the code - similar to how "Ctrl+Click" works on usages of functions and classes
- Generation of an Access Transformer from a class in the IDE. Could have fancy tick boxes for the various transformation options
- Maybe code suggestions for when a non-public method/variable is being used (and by showing an error in the IDE) despite it having a transformer such as when Gradle dependencies haven't been refreshed

DemonWav:
Thanks for the suggestions. Those seem like reasonable features to add.

ZephireNZ:
Oh, and another I thought of: similar to unused methods, highlight unused transformers as well as ones for stuff (such as old versions) that don't exist.

Zidane:
Throwing in my support for this too. Big time saver.

gabizou:
I'm curious how to figure out whether an unused transformer is necessary, because there can be ways that making something public is only resolvable if you scan the entire project for uses in the appropriate level.

DemonWav:
Video showing some of the features that last commit added for access transformers: https://youtu.be/5pJMBYYWPeM

DemonWav:
You think this is finished yet? What other features are missing for this?

DemonWav:
These features are now fully implemented, the last access transformer feature planned is explained in #163.

PR #184, opened 2017-03-17
Allow class, field, and method auto-completion in access transformers

### LINKED ISSUE #138: Code completion for classes, fields and methods in access transformers

DemonWav:
Relevant to #163 and #49
Closes #138 

Brings class auto-completion in the following forms:
  * Package completion. Allows auto-completing package-by-package until the fully qualified class name is built.
  * Short name completion. Allows only entering the short name of the class, such as `WorldServer`, and completing to the fully qualified name, such as `net.minecraft.world.WorldServer`.
  * Anonymous and Inner class completion. Once a base class is entered, typing `$` allows continuing completion to anonymous and inner classes. It supports anonymous class completion, but I don't even know for sure if AT's allow anonymous classes. I assume they do.

Brings field and method auto-completion as well:
  * Field completion works for MCP and non-MCP fields in any class. MCP fields will auto-complete to their SRG name rather than the field typed in. That means the suggestions are based on the MCP name, but the completion uses the SRG name. This automatically adds the standard comment at the end of the line with the MCP name.
  * Method completion works for MCP and non-MCP methods in any class. MCP methods will auto-complete to their SRG name rather than the method typed in. That means the suggestions are based on the MCP name, but the completion uses the SRG name. This automatically adds the standard comment at the end of the line with the MCP name. The suggestions show the the methods' basic signature, except for the return type of the method. The suggestions that are shown only handle method names though, to select a different overload of a method you have to use the arrow keys to select the exact method you mean to transform.

Brings basic auto-completion for keywords:
  * `public`
  * `public-f`
  * `publi+f`
  * `protected`
  * `protected-f`
  * `protected+f`
  * `private`
  * `private-f`
  * `private+f`
  * `default`
  * `default-f`
  * `default+f`


build.gradle.kts
```diff
@@ -216,6 +217,14 @@ val generate = task("generate") {
     outputs.dir("gen")
 }
 
+tasks.withType<RunIdeaTask> {
+    if (System.getProperty("debug") != null) {
+        systemProperty("idea.ProcessCanceledException", "disabled")
+        systemProperty("idea.is.internal", "true")
```
Minecrell: I'm pretty sure this is set automatically.

build.gradle.kts
```diff
@@ -216,6 +217,14 @@ val generate = task("generate") {
     outputs.dir("gen")
 }
 
+tasks.withType<RunIdeaTask> {
```
Minecrell: `runIde.apply {` would be nicer here.

src/main/kotlin/com/demonwav/mcdev/util/McPsiUtil.kt
```diff
@@ -153,3 +153,7 @@ fun PsiType?.isErasureEquivalentTo(other: PsiType?): Boolean {
     // TODO: Do more checks for generics instead
     return TypeConversionUtil.erasure(this) == TypeConversionUtil.erasure(other)
 }
+
+@get:Contract(pure = true)
+val PsiMethod.nameAndParameterTypes: String
+    get() = "$name(${parameterList.parameters.map { it.type.presentableText }.joinToString(", ")})"
```
Minecrell: There is a `transform` parameter in `joinToString` if I remember correctly.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/at/completion/AtCompletionContributor.kt
```diff
@@ -0,0 +1,285 @@
...
+            result.addElement(lookupItem)
+        }
+    }
+
+    private fun handleAtName(memberName: PsiElement, result: CompletionResultSet) {
+        if (memberName !is AtFieldName) {
+            return
+        }
+
+        val text = memberName.text.let { it.substring(0, it.length - 19) }.toLowerCase()
```
Minecrell: Where is 19 coming from? Might be worth extracting it to a constant

src/main/kotlin/com/demonwav/mcdev/platform/mcp/at/completion/SrgPrefixMatcher.kt
```diff
@@ -0,0 +1,27 @@
...
+class SrgPrefixMatcher(prefix: String) : PrefixMatcher(prefix) {
+    override fun prefixMatches(name: String) = true
+    override fun cloneWithPrefix(prefix: String) = SrgPrefixMatcher(prefix)
+
+    override fun prefixMatches(element: LookupElement): Boolean {
+        if (element !is AtMcpLookupItem) {
+            return false
+        }
+
+        return element.getPrettyText().toLowerCase().contains(myPrefix.toLowerCase())
```
Minecrell: `element.getPrettyText().contains(myPrefix, ignoreCase = true)`

src/main/kotlin/com/demonwav/mcdev/platform/mcp/at/completion/AtCompletionContributor.kt
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    private fun handleAtClassName(element: PsiElement, result: CompletionResultSet) {
+        val text = element.text.let { it.substring(0, it.length - 19) }
+        if (text.isEmpty()) {
+            return
+        }
+
+        val currentPackage = text.substringBeforeLast('.')
+        val beginning = text.substringAfterLast('.').toLowerCase()
```
Minecrell: You have to be careful with the default `.toLowerCase()` because it may transform the string incorrectly for some locales. Usually, `.toLowerCase(Locale.ENGLISH)` is what you want to do

src/main/kotlin/com/demonwav/mcdev/platform/mcp/at/completion/AtCompletionContributor.kt
```diff
@@ -0,0 +1,285 @@
...
+        val project = module.project
+
+        // Short name completion
+        if (!text.contains('.')) {
+            val kindResult = result.withPrefixMatcher(KindPrefixMatcher(text))
+            val cache = PsiShortNamesCache.getInstance(project)
+
+            var counter = 0
+            for (className in cache.allClassNames) {
+                if (!className.toLowerCase().startsWith(beginning)) {
```
Minecrell: `startsWith` supports a `ignoreCase` parameter too, you don't need `toLowerCase` necessarily here

build.gradle.kts
```diff
@@ -216,6 +215,13 @@ val generate = task("generate") {
     outputs.dir("gen")
 }
 
+runIde {
```
Minecrell: Can you move this below to `// Use custom JRE ...`. It configures the same task so you could also put in the check for the `intellijJre` property into the configure block. Right now this is in the middle of the code to configure the code generation

PR #186, opened 2017-03-18
Simplify library presentation providers

Minecrell:
The implementation of the various `LibraryPresentationProvider`s is quite similar, so it makes sense to extract the common parts into shared super classes. I think it looks nicer overall, so if you like it just merge it, otherwise feel free to close it. :cat:


src/main/kotlin/com/demonwav/mcdev/platform/bukkit/framework/BukkitPresentationProviders.kt
```diff
@@ -0,0 +1,27 @@
...
+
+class BukkitPresentationProvider : MavenLibraryPresentationProvider(BUKKIT_LIBRARY_KIND, "org.bukkit", "bukkit") {
+    override fun getIcon(properties: LibraryProperties<*>?) = PlatformAssets.BUKKIT_ICON
+}
+
+class PaperPresentationProvider : MavenLibraryPresentationProvider(PAPER_LIBRARY_KIND, "com.destroystokyo.paper", "paper-api") {
+    override fun getIcon(properties: LibraryProperties<*>?) = PlatformAssets.PAPER_ICON
+}
+
+class SpigotPresentationProvider : MavenLibraryPresentationProvider(SPIGOT_LIBRARY_KIND, "org.spigotmc", "spigot-api") {
```
DemonWav: For the sake of ordering, Spigot should be above Paper.

PR #188, opened 2017-03-18
Add code assistance for Mixin configuration files

### LINKED ISSUE #144: Automatically update entry in mixin configuration json when renaming/moving mixin class
I'm not sure if it's already possible, but would be nice to have a way to add these entries automatically, and warn if they are missing (and possibly show warning if class specified in configuration doesn't exist)

Minecrell:
Adds code assistance for Mixin configuration files. Currently it allows completion key names, class/package names and Java compatibility levels.

Closes #144 

The implementation is very similar to #125. The tricky part is how to detect that a JSON file is a Mixin configuration. Currently, I register a custom file type that checks if a file ends with `.json` and starts with `mixins.`. That seems to be a convention that is used in most projects.

screenshot_20170318_190113

screenshot_20170319_172028

Also allows reading the Javadocs using the standard key:

screenshot_20170318_190137

---
### TODO

- [x] Package/class references
- [x] Inspections

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mixin/config/reference/CompatibilityLevel.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mixin/config/reference/CompatibilityLevel.kt
@@ -1,0 +1,61 @@
+/*
+ * Minecraft Dev for IntelliJ
+ *
+ * https://minecraftdev.org
+ *
+ * Copyright (c) 2017 minecraft-dev
+ *
+ * MIT License
+ */
+
+package com.demonwav.mcdev.platform.mixin.config.reference
+
+import com.demonwav.mcdev.platform.mixin.util.MixinConstants.Classes.COMPATIBILITY_LEVEL
+import com.demonwav.mcdev.util.reference.InspectionReference
+import com.intellij.codeInsight.lookup.LookupElementBuilder
+import com.intellij.psi.JavaPsiFacade
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiEnumConstant
+import com.intellij.psi.PsiReference
+import com.intellij.psi.PsiReferenceBase
+import com.intellij.psi.PsiReferenceProvider
+import com.intellij.util.ArrayUtil
+import com.intellij.util.ProcessingContext
+
+object CompatibilityLevel : PsiReferenceProvider() {
+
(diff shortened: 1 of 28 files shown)


src/main/kotlin/com/demonwav/mcdev/platform/mixin/config/reference/CompatibilityLevel.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        override val description
+            get() = "compatibility level '%s'"
+
+        override val unresolved
+            get() = resolve() == null
+
+        override fun resolve(): PsiElement? {
+            val compatibilityLevel = findCompatibilityLevel(element)
+            return compatibilityLevel?.findFieldByName(value, false)?.takeIf { it is PsiEnumConstant }
```
DemonWav: Should this be cached? Seems like the state of this wouldn't change once it's created. Could be easily done using a lazy delegate.

src/main/kotlin/com/demonwav/mcdev/util/reference/ReferenceResolver.kt
```diff
@@ -95,8 +95,8 @@ fun LookupElementBuilder.completeToLiteral(context: PsiElement): LookupElementBu
     // TODO: Currently we replace everything with a single PsiLiteral,
     // not sure how you would keep line breaks after completion
     return withInsertHandler({ context, item ->
-                context.laterRunnable = ReplaceElementWithLiteral(context.editor, context.file, item.lookupString)
-            })
+        context.laterRunnable = ReplaceElementWithLiteral(context.editor, context.file, item.lookupString)
+    })
```
DemonWav: Prefer the `{}` syntax over `({})` where it makes sense and where applicable. This is a spot  where I'd say it makes sense.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/config/reference/CompatibilityLevel.kt
Minecrell: It could change when updating Mixin. I don't think it's very expensive so I'm not sure if caching it is worth it (also need to make sure we don't cause memory leaks by keeping references to invalidated psi elements)

DemonWav: Alright, that's valid.

PR #191, opened 2017-03-23
Properly choose the correct module when creating new Forge mods

DemonWav:
Also: Convert GradleBuildSystem to Kotlin and clean up and organize plugin.xml


DemonWav:
Meh, this is relatively minor. I think it's safe to pull.

PR #193, opened 2017-03-25
Move Gradle tooling extension to separate source set, bundle only Kotlin stdlib

Minecrell:
Currently we have the Gradle tooling extension classes directly in the main source set. However, that has the disadvantage that the full MinecraftDev plugin is added to the Gradle build script classpath when refreshing the Gradle project in IntelliJ IDEA.

We can also not write the tooling extension classes in Kotlin because we can't be sure that Kotlin will be on the buildscript classpath. That will be only the case if the Kotlin IntelliJ IDEA plugin is applied.

I've moved it to a separate source set that gets built to a separate JAR, so we add only the necessary classes to the buildscript classpath. There is an additional commit that removes `kotlin-runtime` and `kotlin-stdlib` from the bundled dependencies since these are already included in the IntelliJ IDEA distribution.


gradle.properties
```diff
@@ -8,7 +8,7 @@
 # MIT License
 #
 # suppress inspection "UnusedProperty" for whole file
-ideaVersion = IC-2017.1
+ideaVersion = 2017.1
```
DemonWav: ?

Minecrell: https://github.com/minecraft-dev/MinecraftDev/pull/193/files#diff-392475fdf2bc320d17762ed97109a121R108

PR #197, opened 2017-04-01
Add MixinEntryPoint to suggest lower private visibility for Mixin injectors

Minecrell:
Currently, the `ImplicitUsageProvider` added in https://github.com/minecraft-dev/MinecraftDev/commit/7e2fe99578ec6f231bc4441683f4dbaa1f809a09 (as well as the previous way) break the `Declaration access can be weaker` inspection. Mixin injectors can be usually private so it would be good to have the existing inspection working (because it handles all the cases in which  they can't be private).

I've moved the entry points from `ImplicitUsageProvider` to a custom `EntryPoint` implementation that declares the minimal visibility for Mixin injectors as private:

screenshot_20170401_151933


src/main/kotlin/com/demonwav/mcdev/platform/mixin/insight/MixinEntryPoint.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    override fun isEntryPoint(refElement: RefElement, psiElement: PsiElement) = isEntryPoint(psiElement)
+
+    override fun getMinVisibilityLevel(member: PsiMember): Int {
+        if (member !is PsiMethod) return -1
+        val modifierList = member.modifierList
+        return if (MixinConstants.Annotations.METHOD_INJECTORS.any {
+            modifierList.findAnnotation(it) != null
+        }) PsiUtil.ACCESS_LEVEL_PRIVATE else -1
```
DemonWav: Only omit curly braces when the whole statement is written on one line (i.e., is effectively a ternary). Here the  conditional spans 3 lines, so the whole statement doesn't fit on one line.

PR #213, opened 2017-05-15
Override the JavaModuleType display name and tooltip in MinecraftModuleType.

JBYoshi:
The tooltip for the Minecraft option on the project creation wizard currently shows this:
`Java modules are used for developing JVM-based desktop and web applications...`
This replaces it with a Minecraft-specific description: `Minecraft modules are used for developing plugins or mods for <b>Minecraft</b> (Java Edition, also known as the PC Edition).` It also overrides the related `getName()` method I saw to prevent possible questions about that in the future.

(I specifically mentioned the Java edition because the future Pocket Edition plugins will use C#, not Java, so they will probably use a significantly different module type.)

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/MinecraftModuleType.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/MinecraftModuleType.kt
@@ -25,9 +25,9 @@ class MinecraftModuleType : JavaModuleType() {
     override fun getDescription() = "Minecraft modules are used for developing plugins or mods for <b>Minecraft</b> (Java Edition, also known as the PC Edition)."
 
     companion object {
-        private val ID = "MINECRAFT_MODULE_TYPE"
-        val NAME = "Minecraft"
-        val OPTION = "com.demonwav.mcdev.MinecraftModuleTypes"
+        private const val ID = "MINECRAFT_MODULE_TYPE"
+        const val NAME = "Minecraft"
+        const val OPTION = "com.demonwav.mcdev.MinecraftModuleTypes"
 
         val instance: MinecraftModuleType
             get() = ModuleTypeManager.getInstance().findByID(ID) as MinecraftModuleType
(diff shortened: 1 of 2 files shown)


DemonWav:
@JBYoshi errors on this PR:

```
e: /home/ubuntu/MinecraftDev/src/main/kotlin/com/demonwav/mcdev/creator/MinecraftModuleBuilder.kt: (38, 41): Unresolved reference: MinecraftModuleType
e: /home/ubuntu/MinecraftDev/src/main/kotlin/com/demonwav/mcdev/creator/MinecraftModuleBuilder.kt: (40, 35): Unresolved reference: MinecraftModuleType
```

src/main/kotlin/com/demonwav/mcdev/platform/MinecraftModuleType.kt
```diff
@@ -21,9 +21,12 @@ class MinecraftModuleType : JavaModuleType() {
     override fun getBigIcon() = PlatformAssets.MINECRAFT_ICON_2X
     override fun getIcon() = PlatformAssets.MINECRAFT_ICON
     override fun getNodeIcon(isOpened: Boolean) = PlatformAssets.MINECRAFT_ICON
+    override fun getName() = NAME
+    override fun getDescription() = "Minecraft modules are used for developing plugins or mods for <b>Minecraft</b> (Java Edition, also known as the PC Edition)."
 
     companion object {
         private val ID = "MINECRAFT_MODULE_TYPE"
+        val NAME = "Minecraft"
         val OPTION = "com.demonwav.mcdev.MinecraftModuleTypes"
```
DemonWav: Go ahead and do me a solid and make all three of these `const` as well.

JBYoshi: `ID.const = true`
`NAME.const = true`
`OPTION.const = true`
Done :wink:

PR #225, opened 2017-05-30
Forge only wants lowercased MODIDs, let's sanity check that!

cpw:
In line with vanilla since 1.11, modids should only be lowercase. This PR attempts to (no doubt awfully) try and enforce that restriction for Forge mods.

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/creator/ForgeProjectSettingsWizard.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/creator/ForgeProjectSettingsWizard.kt
@@ -10,6 +10,7 @@
 
 package com.demonwav.mcdev.creator
 
+import com.demonwav.mcdev.exception.MinecraftSetupException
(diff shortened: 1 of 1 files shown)


DenWav:
In this case perhaps the creator should have separate fields for mod name and mod id, right? Since mod name wouldn't also have this restriction.

cpw:
The mod's name isn't generally the modid no. The modid has some special rules. All lowercase is the biggest one. It gets transformed into a file path which means no colons. It also has a length restriction.

stephan-gh:
The same applies to Sponge so it would be good to have the separate fields for name + ID and the check there as well.

DenWav:
Do you want to try and fiddle with the UI to accomplish that @cpw? Feel free if you do. If not, it won't be difficult for me to implement this.

cpw:
I think you know your way around the UI better than I do @DemonWav.

DenWav:
I've created issue #339 to address this, as I'll be making changes to the UI to fix this.

PR #238, opened 2017-06-14
Add code folding for Mixin injection targets

stephan-gh:
Adds some code folding for Mixin classes. Right now there is an option to fold target descriptors and the (in my opinion) annoying `(Object)` casts that are necessary sometimes.

Both options can be toggled in the Code folding options. As we discussed a while ago, I've made the folding of target descriptors enabled by default and disabled the folding of object casts.

<details> 
  <summary>Screenshot of target descriptor folding</summary>

screenshot_20170614_170201

</details>

<details> 
  <summary>Screenshot of Object cast folding</summary>

screenshot_20170614_170341

</details>

<p><br>The original code will appear when clicking on the folded code, just like when IntelliJ IDEA collapses method bodies etc.</p>

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mixin/folding/MixinFoldingBuilder.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mixin/folding/MixinFoldingBuilder.kt
@@ -1,0 +1,124 @@
+/*
+ * Minecraft Dev for IntelliJ
+ *
+ * https://minecraftdev.org
+ *
+ * Copyright (c) 2017 minecraft-dev
+ *
+ * MIT License
+ */
+
+package com.demonwav.mcdev.platform.mixin.folding
+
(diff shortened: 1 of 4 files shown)


gabizou:
Would you be able to fold the double casts? Or maybe even include some smart double cast options when typing `this` in a mixin? Might need another pr or feature set addition for the second part.

stephan-gh:
@gabizou Are you referring to casts like `(EntityLivingBase) (Object) this`? Folding casts entirely would only make sense in that case IMO (and only if it actually matches the target class)

kashike:
@gabizou https://korobi.io/network/esper/channel/mcdev/logs/2017/04/12/#L19-36

gabizou:
@kashike It still is mildly annoying where it does show it as highlighted as though it is unnecessary, but it would still be worthwhile to have some sort of folding to increase legibility in the target code.

DenWav:
Looks  good, thanks as always. Sorry for the long wait. :heart:

PR #272, opened 2017-07-20
Fix Forge project generation for MC1.12+ and enhance Minecraft version selection

marvin-roesch:
As of version 1.12 of Minecraft, Forge projects are required to use ForgeGradle 2.3, while earlier versions should still use FG 2.2. This was not accounted for in the plugin, but this PR fixes the issue by making the ForgeGradle version variable based on the Minecraft version. Since I implemented a `SemanticVersion` utility for this which allows us to compare Minecraft (and Forge, technically) versions easily, I also "fixed" the way the initial Minecraft version is retrieved in the wizard, namely by properly comparing the versions, assuming semantic versioning. The previous method of just comparing version strings was (currently) always yielding "1.9.4", since the String comparison compares each character individually and favours those that come later in the charset. This did not respect the components of the semantic versions. 1.9.4 was alwyas selected because the '9' was always the biggest character at that position, so we wouldn't get another result until Minecraft 1.90 is released :D.

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/util/SemanticVersion.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/util/SemanticVersion.kt
@@ -13,18 +13,18 @@ package com.demonwav.mcdev.util
 import com.demonwav.mcdev.util.SemanticVersion.Companion.VersionPart.PreReleasePart
 import com.demonwav.mcdev.util.SemanticVersion.Companion.VersionPart.ReleasePart
 
-class SemanticVersion(val value: String) : Comparable<SemanticVersion> {
-    val parts = value.split(".").map {
-        if (it.contains("_pre")) {
-            val subParts = it.split("_pre")
-            PreReleasePart(subParts[0].toInt(), subParts[1].toInt())
-        } else {
-            ReleasePart(it.toInt())
-        }
-    }
-
+/**
+ * Represents a comparable and generalised "semantic version".
+ * Each constituent part (delimited by periods in a version string) contributes
+ * to the version ranking with decreasing priority from left to right.
+ */
+class SemanticVersion(val parts: List<VersionPart>) : Comparable<SemanticVersion> {
     override fun compareTo(other: SemanticVersion): Int {
-        val result = parts.zip(other.parts).fold(0, { acc, (a, b) -> if (acc == -1) acc else a.compareTo(b) })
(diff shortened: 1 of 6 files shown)


kashike (approved):

src/main/kotlin/com/demonwav/mcdev/util/SemanticVersion.kt
```diff
@@ -0,0 +1,63 @@
...
+ * MIT License
+ */
+
+package com.demonwav.mcdev.util
+
+import com.demonwav.mcdev.util.SemanticVersion.Companion.VersionPart.PreReleasePart
+import com.demonwav.mcdev.util.SemanticVersion.Companion.VersionPart.ReleasePart
+
+class SemanticVersion(val value: String) : Comparable<SemanticVersion> {
+    val parts = value.split(".").map {
```
DenWav: This should be `.split('.')` (split on the char rather than the String).

src/main/kotlin/com/demonwav/mcdev/util/SemanticVersion.kt
```diff
@@ -0,0 +1,63 @@
...
+        if (it.contains("_pre")) {
+            val subParts = it.split("_pre")
+            PreReleasePart(subParts[0].toInt(), subParts[1].toInt())
```
DenWav: How confident are we in the format of these versions that this will never have index out of bounds issues?

src/main/kotlin/com/demonwav/mcdev/util/SemanticVersion.kt
```diff
@@ -0,0 +1,63 @@
...
+        } else {
+            ReleasePart(it.toInt())
+        }
+    }
+
+    override fun compareTo(other: SemanticVersion): Int {
+        val result = parts.zip(other.parts).fold(0, { acc, (a, b) -> if (acc == -1) acc else a.compareTo(b) })
```
DenWav: This should use function syntax:

`.fold(0) { ... }`

DenWav (changes requested):

src/main/kotlin/com/demonwav/mcdev/util/SemanticVersion.kt
marvin-roesch: In the context of MC versions, we can be 100% confident, I think, but I'll add some safeguards to prevent more general purpose usage blowing up.

DenWav (approved):

PR #279, opened 2017-07-31
Update readme.md

XorNick:


DemonWav:
Is it not still true that it is the _only_ plugin that shows up?

nickeh:
@DemonWav It's still true, it's just a precaution for the future.

XorNick:
meh?

PR #303, opened 2017-09-12
Minor cleanup everywhere

stephan-gh:
I think this should be pretty straightforward. No functional change. Just some minor cleanup to fix some warnings reported by IntelliJ inspections (but not all of them, because some don't make sense). Opening this pull request if you want to take a quick look @DemonWav.

- Fixed wrong package declaration
- Replaced some long if-else chains with when
- Moved some assignments / return statements outside of if/try blocks
- Other minor cleanup to use more Kotlin features

I avoided touching any Forge side-only inspections since you have a branch somewhere that removes/changes them.

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/forge/version/ForgeVersion.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/forge/version/ForgeVersion.kt
@@ -10,8 +10,8 @@
 
 package com.demonwav.mcdev.platform.forge.version
 
-import com.demonwav.mcdev.util.gson
 import com.demonwav.mcdev.util.SemanticVersion
+import com.demonwav.mcdev.util.gson
 import com.demonwav.mcdev.util.sortVersions
 import java.io.IOException
 import java.net.URL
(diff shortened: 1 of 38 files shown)


marvin-roesch:
You might want to address some occurrences of `string.trim { it <= ' ' }` which seem to be the result of automatic conversion from Java.

stephan-gh:
@PaleoCrafter Thanks, done.

src/main/kotlin/com/demonwav/mcdev/platform/forge/version/ForgeVersion.kt
```diff
@@ -10,8 +10,8 @@
 
 package com.demonwav.mcdev.platform.forge.version
 
-import com.demonwav.mcdev.util.gson
 import com.demonwav.mcdev.util.SemanticVersion
+import com.demonwav.mcdev.util.gson
```
DenWav, line 14: This seems weird to me. Does uppercase sort before lowercase? Seems like imports should be sorted case-insensitive to me.

stephan-gh: That's what IntelliJ IDEA does when sorting the imports. It seems like they are sorted case-sensitive, but since this is what your IntelliJ code style does, I'm not sure we should bother and keep this different manually

PR #308, opened 2017-10-01
Touchup Canary Platform

jamierocks:
- Don't refer to hooks as 'event's in parameter names
- Add icon to Neptune.inf files (optional plugin descriptor in NeptuneMod)
- Reinstate the MCP text format colouring (although for both canary + neptune this time)

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/canary/CanaryModule.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/canary/CanaryModule.kt
@@ -32,6 +32,7 @@ class CanaryModule<out T : AbstractModuleType<*>>(facet: MinecraftFacet, overrid
 
     override val type: PlatformType = moduleType.platformType
     private var canaryInf: VirtualFile? = null
+    private var neptuneInf: VirtualFile? = null
 
     init {
         setup()
@@ -39,6 +40,7 @@ class CanaryModule<out T : AbstractModuleType<*>>(facet: MinecraftFacet, overrid
 
     private fun setup() {
         canaryInf = facet.findFile(CanaryConstants.CANARY_INF, SourceType.RESOURCE)
+        neptuneInf = facet.findFile(CanaryConstants.NEPTUNE_INF, SourceType.RESOURCE)
     }
 
     fun getCanaryInf(): VirtualFile? {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/demonwav/mcdev/platform/canary/CanaryModule.kt
```diff
@@ -50,6 +52,15 @@ class CanaryModule<out T : AbstractModuleType<*>>(facet: MinecraftFacet, overrid
         return canaryInf
     }
 
+    fun getNeptuneInf(): VirtualFile? {
```
stephan-gh: You can implement this as regular property (using a custom getter and a private setter)

Same for the `canaryInf` above

stephan-gh (changes requested):

src/main/kotlin/com/demonwav/mcdev/platform/canary/CanaryModule.kt
```diff
@@ -31,23 +31,34 @@ import com.intellij.psi.search.GlobalSearchScope
 class CanaryModule<out T : AbstractModuleType<*>>(facet: MinecraftFacet, override val moduleType: T) : AbstractModule(facet) {
 
     override val type: PlatformType = moduleType.platformType
-    private var canaryInf: VirtualFile? = null
+    var canaryInf: VirtualFile? = null
+        get() {
+            if (field == null) {
```
stephan-gh: Another thought: Do the null checks for `canaryInf` and `neptuneInf` in `setup()` (and initialize them if needed) and then call `setup()` every time. That way we can reduce duplication...

stephan-gh (changes requested):

src/main/kotlin/com/demonwav/mcdev/platform/canary/CanaryModule.kt
jamierocks: huh, thought I'd already done that :S

stephan-gh (approved):
🎉

kashike (approved):

DenWav:
I have an idea that I think will cleanup your lazy values, but I can see it being useful for every platform, so I will implement it after I merge.

PR #315, opened 2017-10-14
[WIP] Introduce JSON validation through schemas

marvin-roesch:
As a disclaimer, this is very much to be interpreted as a WIP feature right now, but I still wanted to get out the current results and iteratively improve upon them in order to get early feedback from people manually using the schemas.

This PR makes the first steps towards providing helpful support for the various proprietary JSON "formats" put forward by Mojang and other parties (such as Forge). Currently, the following formats are being worked on or finished (the paths correlate directly to what is matched by the schemas):
- [x] `sounds.json` (first thing finished because it's a really simple format)
- [x] `blockstates/[file].json` (both Vanilla and Forge flavours)
- [x] `models/(item/block)/[file].json`
- [ ] `recipes/**/[file].json` (only Vanilla and Forge types for now)
- [ ] `advancements/**/[file].json`
- [x] `loot_tables/**/[file].json`

Some early feedback would be appreciated and if somebody is familiar with the [JSON Schema format](http://json-schema.org/), suggestions on how to improve mine would be great. Unfortunately, IntelliJ's built-in validator appears to be fairly fenced off code-wise, the only entrypoint I could find was the schema providers to add custom schemas, with no way to provide custom error messages of any kind.

This limitation may be addressed on a second PR I plan to make (which may or may not be worked on in parallel with this, depending on how easy it is), which aims to provide intelligent support for JSON editing, such as file referencing and (limited) registration validation.


Ayfri:
Recipes JSON are now also usable for Fabric, should update.

TheOnlyTails:
I'm familiar with JSON schemas, I'll try and help.

Also opened an issue here: #1049

PR #318, opened 2017-11-11
Ungrab mouse when pausing while debugging

### LINKED ISSUE #132: Add option to automatically ungrab mouse
I suggest adding option that will ungrab mouse automatically when reaching breakpoint. At the moment I do this steps every time I'm using breakpoints:
1. Press `Alt+F8`
2. Type `Mouse.setGrabbed(false)`
3. Press Enter
4. Press Escape

I think this option will be useful because doing this steps takes much time.

stephan-gh:
Closes #132 

Apparently working correctly according to @kashike and @afdw. Code is slightly hacky but it does the job...

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mcp/debug/McpRunConfigurationExtension.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mcp/debug/McpRunConfigurationExtension.kt
@@ -1,0 +1,43 @@
+/*
+ * Minecraft Dev for IntelliJ
+ *
+ * https://minecraftdev.org
+ *
+ * Copyright (c) 2017 minecraft-dev
+ *
+ * MIT License
+ */
+
+package com.demonwav.mcdev.platform.mcp.debug
+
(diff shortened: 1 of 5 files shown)


realmayus:
This doesn't fix the issue for me, **I'm on Pop!_OS 19.10**
IntelliJ "about" info:
```
IntelliJ IDEA 2019.3.3 (Ultimate Edition)
Build #IU-193.6494.35, built on February 11, 2020
Licensed to 
Subscription is active until January 22, 2021
For educational use only.
Runtime version: 11.0.5+10-b520.38 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Linux 5.3.0-7629-generic
GC: ParNew, ConcurrentMarkSweep
Memory: 1925M
Cores: 12
Registry: 
Non-Bundled Plugins: com.demonwav.minecraft-dev
```

DenWav:
Newer versions of Minecraft use a new version of LWJGL which isn't compatible with this change - #688 fixes this.

PR #320, opened 2017-11-20
add copy AT entry menu item

thiakil:
Basically a copy of the Get SRG Name task, which copies a (partial) AT entry to the clipboard. Only missing the access specifier.

Currently to do this I need to use the Get SRG Name action, select the text in the balloon and `ctrl-c` (and enter the rest manually of course)

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
@@ -14,8 +14,6 @@ import com.intellij.openapi.wm.WindowManager
 import com.intellij.psi.*
 import com.intellij.ui.LightColors
 import com.intellij.ui.awt.RelativePoint
-import java.awt.Toolkit.getDefaultToolkit
-import javafx.scene.input.Clipboard.getSystemClipboard
 import java.awt.Toolkit
 import java.awt.datatransfer.StringSelection
 
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
@@ -1,0 +1,106 @@
+package com.demonwav.mcdev.platform.mcp.actions
+
+import com.demonwav.mcdev.platform.mcp.McpModuleType
+import com.demonwav.mcdev.platform.mixin.util.findFirstShadowTarget
+import com.demonwav.mcdev.util.getDataFromActionEvent
+import com.demonwav.mcdev.util.invokeLater
+import com.intellij.openapi.actionSystem.AnAction
+import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.actionSystem.DataKeys
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.ui.popup.Balloon
+import com.intellij.openapi.ui.popup.JBPopupFactory
+import com.intellij.openapi.wm.WindowManager
+import com.intellij.psi.*
+import com.intellij.ui.LightColors
+import com.intellij.ui.awt.RelativePoint
+import java.awt.Toolkit.getDefaultToolkit
+import javafx.scene.input.Clipboard.getSystemClipboard
+import java.awt.Toolkit
+import java.awt.datatransfer.StringSelection
+
+
+
+class CopyATAction : AnAction() {
+    override fun actionPerformed(e: AnActionEvent) {
+        val data = getDataFromActionEvent(e) ?: return showBalloon("Unknown failure", e)
+
+        if (data.element !is PsiIdentifier) {
+            showBalloon("Not a valid element", e)
+            return
+        }
+
+        val mcpModule = data.instance.getModuleOfType(McpModuleType) ?: return showBalloon("No mappings found", e)
+
+        mcpModule.srgManager?.srgMap?.done { srgMap ->
+            var parent = data.element.parent
+
+            if (parent is PsiMember) {
+                val shadowTarget = parent.findFirstShadowTarget()
+                if (shadowTarget != null) {
+                    parent = shadowTarget
+                }
+            }
+
+            if (parent is PsiReference) {
+                parent = parent.resolve()
+            }
+
+            when (parent) {
+                is PsiField -> {
+                    val srg = srgMap.findSrgField(parent) ?: return@done showBalloon("No SRG name found", e)
+                    copyToClipboard(data.editor, data.element, parent.containingClass?.qualifiedName + " " + srg.name + " #"+parent.name)
+                }
+                is PsiMethod -> {
+                    val srg = srgMap.findSrgMethod(parent) ?: return@done showBalloon("No SRG name found", e)
+                    copyToClipboard(data.editor, data.element, parent.containingClass?.qualifiedName + " " + srg.name + srg.descriptor+ " #"+parent.name)
+                }
+                is PsiClass -> {
+                    val classMcpToSrg = srgMap.findSrgClass(parent) ?: return@done showBalloon("No SRG name found", e)
+                    copyToClipboard(data.editor, data.element, classMcpToSrg)
+                }
+                else ->
+                    showBalloon("Not a valid element", e)
+            }
+        } ?: showBalloon("No mappings found", e)
+    }
+
+    private fun showBalloon(message: String, e: AnActionEvent) {
+        val balloon = JBPopupFactory.getInstance()
+            .createHtmlTextBalloonBuilder(message, null, LightColors.YELLOW, null)
+            .setHideOnAction(true)
+            .setHideOnClickOutside(true)
+            .setHideOnKeyOutside(true)
+            .createBalloon()
+
+        val statusBar = WindowManager.getInstance().getStatusBar(DataKeys.PROJECT.getData(e.dataContext))
+
+        invokeLater { balloon.show(RelativePoint.getCenterOf(statusBar.component), Balloon.Position.atRight) }
+    }
+
+    private fun copyToClipboard(editor: Editor, element: PsiElement, text: String) {
+        val stringSelection = StringSelection(text)
+        val clpbrd = Toolkit.getDefaultToolkit().systemClipboard
+        clpbrd.setContents(stringSelection, null)
+        showSuccessBalloon(editor, element, "Copied "+text)
+    }
+
+    private fun showSuccessBalloon(editor: Editor, element: PsiElement, text: String) {
+        val balloon = JBPopupFactory.getInstance()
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            if (parent is PsiReference) {
+                parent = parent.resolve()
+            }
+
+            when (parent) {
+                is PsiField -> {
+                    val srg = srgMap.findSrgField(parent) ?: return@done showBalloon("No SRG name found", e)
+                    copyToClipboard(data.editor, data.element, parent.containingClass?.qualifiedName + " " + srg.name + " #"+parent.name)
```
DenWav: No space before and after `+`.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
```diff
@@ -0,0 +1,114 @@
...
+                }
+                is PsiMethod -> {
+                    val srg = srgMap.findSrgMethod(parent) ?: return@done showBalloon("No SRG name found", e)
+                    copyToClipboard(data.editor, data.element, parent.containingClass?.qualifiedName + " " + srg.name + srg.descriptor+ " #"+parent.name)
```
DenWav: No space before and after `+`.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
```diff
@@ -0,0 +1,114 @@
...
+                }
+                is PsiClass -> {
+                    val classMcpToSrg = srgMap.findSrgClass(parent) ?: return@done showBalloon("No SRG name found", e)
+                    copyToClipboard(data.editor, data.element, classMcpToSrg)
+                }
+                else ->
+                    showBalloon("Not a valid element", e)
```
DenWav: Place this on the same line as the `else`.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
```diff
@@ -0,0 +1,114 @@
...
+        val balloon = JBPopupFactory.getInstance()
+            .createHtmlTextBalloonBuilder(message, null, LightColors.YELLOW, null)
+            .setHideOnAction(true)
+            .setHideOnClickOutside(true)
+            .setHideOnKeyOutside(true)
+            .createBalloon()
+
+        val statusBar = WindowManager.getInstance().getStatusBar(DataKeys.PROJECT.getData(e.dataContext))
+
+        invokeLater { balloon.show(RelativePoint.getCenterOf(statusBar.component), Balloon.Position.atRight) }
```
DenWav: Don't 1-line this. Place the contents of `{ ... }` on a new line.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
```diff
@@ -0,0 +1,114 @@
...
+            balloon.show(
+                    RelativePoint(
+                            editor.contentComponent,
+                            editor.visualPositionToXY(editor.offsetToVisualPosition(element.textRange.endOffset))
+                    ),
+                    Balloon.Position.atRight
+            )
+        }
+    }
+}
```
DenWav: Needs a new-line at the end of the file.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
```diff
@@ -0,0 +1,114 @@
...
+            .setHideOnKeyOutside(true)
+            .createBalloon()
+
+        invokeLater {
+            balloon.show(
+                    RelativePoint(
+                            editor.contentComponent,
+                            editor.visualPositionToXY(editor.offsetToVisualPosition(element.textRange.endOffset))
+                    ),
+                    Balloon.Position.atRight
```
DenWav: Should be a 4-space indent, not 8.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
```diff
@@ -0,0 +1,114 @@
...
+import com.intellij.openapi.wm.WindowManager
+import com.intellij.psi.*
+import com.intellij.ui.LightColors
+import com.intellij.ui.awt.RelativePoint
+import java.awt.Toolkit
+import java.awt.datatransfer.StringSelection
+
+
+
+class CopyATAction : AnAction() {
```
DenWav: Should only be 1 new-line between imports and class.

DenWav (changes requested):
I like the idea, implementation seems mostly fine, but after a quick glance, you need to fix some styling issues.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
thiakil: That's actually a direct copy from `FindSrgMappingAction`

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
thiakil: That's actually a direct copy from `FindSrgMappingAction`

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
DenWav: Thanks, I'll fix it there too.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
```diff
@@ -0,0 +1,113 @@
...
+import com.intellij.openapi.ui.popup.Balloon
+import com.intellij.openapi.ui.popup.JBPopupFactory
+import com.intellij.openapi.wm.WindowManager
+import com.intellij.psi.*
+import com.intellij.ui.LightColors
+import com.intellij.ui.awt.RelativePoint
+import java.awt.Toolkit
+import java.awt.datatransfer.StringSelection
+
+class CopyATAction : AnAction() {
```
stephan-gh: I know it's weird, but the way we name stuff this should be `CopyAtAction`.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
```diff
@@ -0,0 +1,113 @@
...
+        val data = getDataFromActionEvent(e) ?: return showBalloon("Unknown failure", e)
+
+        if (data.element !is PsiIdentifier) {
+            showBalloon("Not a valid element", e)
+            return
+        }
+
+        val mcpModule = data.instance.getModuleOfType(McpModuleType) ?: return showBalloon("No mappings found", e)
+
+        mcpModule.srgManager?.srgMap?.done { srgMap ->
```
stephan-gh: Can we avoid duplicating this part from `FindSrgMappingAction`? It should be possible to move this to a separate method since the code appears to be almost identical.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
```diff
@@ -0,0 +1,113 @@
...
+                is PsiClass -> {
+                    val classMcpToSrg = srgMap.findSrgClass(parent) ?: return@done showBalloon("No SRG name found", e)
+                    copyToClipboard(data.editor, data.element, classMcpToSrg)
+                }
+                else -> showBalloon("Not a valid element", e)
+            }
+        } ?: showBalloon("No mappings found", e)
+    }
+
+    private fun showBalloon(message: String, e: AnActionEvent) {
```
stephan-gh: As above, move this to an utility method that is shared between both actions.

stephan-gh (changes requested):

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
DenWav: Thanks for catching stuff. It's not weird though, it has it's reasons 😛.

DenWav (dismissed):

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
thiakil: Certainly. What would be the best way to do this in Kotlin, my first assumption would be to make them have a common superclass which includes the common functions?

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
thiakil: Well I left it like that since it's an acronym :P

Would `CopyAccessTransformerAction` be better?

DenWav: No. `CopyAtAction` is correct. [The reasoning is explained in the style guide.](https://www.demonwav.com/style#s5.3-camel-case)

thiakil: I get that it's correct, but using the non abbreviated one potentially makes us both happy :p

DenWav: I see. No, please call it `CopyAtAction`. Within the context of this code it is clear what that means, no need to write out the `AccessTransformers` bit. I think writing it out looks worse.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/actions/CopyATAction.kt
DenWav: I haven't looked into the exact details regarding your implementation, unfortunately due to lack of time. But depending on the context of the situation, either a common super class, or top level functions would work well. Intuition regarding what you're working on tells me a commons super class is likely the cleanest solution.

If you want top level functions instead, and you want to namespace the functions a bit you can place them in an `object`. Unfortunately Kotlin doesn't have `package-private` visibility to help clean up the global namespace, so sometimes `object`s are the best way to keep it clean

thiakil:
Have merged the common functions, though not sure if that's what you had in mind

PR #352, opened 2018-02-01
Auto-complete keys in localizations from default

### LINKED ISSUE #347: Auto-complete lang keys based on keys present in the default lang file

marvin-roesch:
This resolves #347. The auto-completion will suggest keys not yet in a localization that exist in the default language file (of that domain).

I mostly modelled this after the auto-completion provider for ATs, so similarities might be noticeable :P


DenWav (approved):
You're getting better with the formatting.

PR #378, opened 2018-04-23
Add import optimizer for Mixin configuration files

### LINKED ISSUE #377: Allow sorting mixin entries in mixin config file
Projects like Sponge try to keep the mixins in the `mixins` section of a Mixin config sorted in alphabetical order. It would be nice if MinecraftDev provided an option to automatically sort this section of a Mixin config.

stephan-gh:
Closes #377

Implements a custom `ImportOptimizer` for Mixin configuration files that sorts the Mixins listed in the Mixin configuration file (and removes duplicates eventually).

Integrating this as the standard `Optimize imports` IntelliJ feature makes the implementation easier (no custom actions etc). To some extend, the Mixins listed in the configuration represent some kind of imports so I think it makes sense. That way it also can be applied to entire directory, automatically on commit etc etc

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mixin/config/MixinConfigImportOptimizer.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mixin/config/MixinConfigImportOptimizer.kt
@@ -1,0 +1,130 @@
+/*
+ * Minecraft Dev for IntelliJ
+ *
+ * https://minecraftdev.org
+ *
+ * Copyright (c) 2018 minecraft-dev
+ *
+ * MIT License
+ */
+
+package com.demonwav.mcdev.platform.mixin.config
+
+import com.google.common.collect.Comparators
+import com.google.gson.Gson
+import com.intellij.json.psi.JsonArray
+import com.intellij.json.psi.JsonElementGenerator
+import com.intellij.json.psi.JsonFile
+import com.intellij.json.psi.JsonObject
+import com.intellij.json.psi.JsonProperty
+import com.intellij.json.psi.JsonStringLiteral
+import com.intellij.lang.ImportOptimizer
+import com.intellij.openapi.util.EmptyRunnable
+import com.intellij.psi.PsiDocumentManager
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+
+class MixinConfigImportOptimizer : ImportOptimizer {
+
+    /**
+     * Sorts a number of (qualified) class names based on their package names
+     * and the following rules:
+     *
+     *   - Each package name is compared separately
+     *   - Default packages come first
+     *   - Class names are only compared if in the same package
+     *
+     * Example:
+     *   ```
+     *   Main
+     *   com.demonwav.mcdev.Example
+     *   com.demonwav.mcdev.HelloWorld
+     *   com.demonwav.mcdev.asset.Assets
+     *   com.demonwav.mcdev.util.ActionData
+     *   ```
+     */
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/demonwav/mcdev/platform/mixin/config/MixinConfigImportOptimizer.kt
```diff
@@ -0,0 +1,130 @@
...
+     *
+     * Example:
+     *   ```
+     *   Main
+     *   com.demonwav.mcdev.Example
+     *   com.demonwav.mcdev.HelloWorld
+     *   com.demonwav.mcdev.asset.Assets
+     *   com.demonwav.mcdev.util.ActionData
+     *   ```
+     */
```
DenWav, line 45: As long as this is the sort that's expected from this feature, that's fine by me,  though it doesn't follow IntelliJ's default sorting behavior.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/config/MixinConfigImportOptimizer.kt
```diff
@@ -0,0 +1,130 @@
...
+                // Default package always comes first
+                return Integer.compare(parts1.size, parts2.size)
+            }
+
+            // Compare class names
+            return parts1[end].compareTo(parts2[end], ignoreCase = true)
+        }
+    }
+
+    private val gson = Gson()
```
DenWav, line 73: There's a global `gson` object in [`utils.kt`](https://github.com/minecraft-dev/MinecraftDev/blob/f2e99b667210987e77b16daea13a5aa6a2e0a9af/src/main/kotlin/com/demonwav/mcdev/util/utils.kt#L150-L151) you should use instead.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/config/MixinConfigImportOptimizer.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    private fun processMixins(property: JsonProperty?): JsonArray? {
+        property ?: return null
+        val mixins = property.value as? JsonArray ?: return null
+        val classes = mixins.valueList.mapNotNull { (it as? JsonStringLiteral)?.value }
+        if (Comparators.isInStrictOrder(classes, ClassPackageComparator)) {
+            return null
+        }
+
+        // Kind of lazy here, serialize the sorted list and let IntelliJ parse it
```
DenWav, line 103: Now this is what I call lazy, but I don't have an issue with it.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/config/MixinConfigImportOptimizer.kt
```diff
@@ -0,0 +1,130 @@
...
+        if (mixins != null || server != null || client != null) {
+            return Task(file, mixins, server, client)
+        }
+
+        return EmptyRunnable.getInstance()
+    }
+
+    private fun processMixins(property: JsonProperty?): JsonArray? {
+        property ?: return null
+        val mixins = property.value as? JsonArray ?: return null
```
DenWav, line 97: Could just do `val mixins = property?.value as? JsonArray ?: return null` instead of the parameter check in it's own line.

DenWav (changes requested):
Not any major things, everything looks pretty straight forward. Just make the couple of pointed out changes and it'll be good to merge.

PR #384, opened 2018-05-10
Suppress the ClassCastException warnings in Mixins

### LINKED ISSUE #374: Add ignore for ClassCastException Warning for Mixins
Target Platforms: Mixin

In short, IntelliJ 2018.1 introduced ClassCastException detection warnings and of course, Mixins use `(targetClass) (Object) this` to cast back and at runtime those casts are eliminated, but IntelliJ will complain regardless.

Providing a screenshot of the warning in an example:
screen shot 2018-04-08 at 12 21 32 pm

DenWav:
This applies to the common practice in Mixins to do
`(SomeClass) (Object) this`, which IntelliJ now (if these weren't
Mixins, correctly) warns that this expression might cause a CCE. For
these scenarios, when the cast types match correctly enough with Mixin
targets, this will suppress that inspection.

Fixes #374

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mixin/inspection/suppress/MixinClassCastInspectionSuppressor.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mixin/inspection/suppress/MixinClassCastInspectionSuppressor.kt
@@ -47,9 +47,14 @@ class MixinClassCastInspectionSuppressor : InspectionSuppressor {
         val factory = JavaPsiFacade.getInstance(project).elementFactory
 
         val toType = castType.type
-        if (!targets.asSequence()
-            .map(factory::createType)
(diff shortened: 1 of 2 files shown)


Niko-sk2x:
Would it also be possible to automatically suggest this cast as a fix for compile error when trying to use `this` as object of the mixin target class?

Intellij already seems to be able to apply such fix automatically in case like this (this is just a nonsensical example to show that IDEA can already do such thing when it knows the object is of that type):
```java
    public World getWorld() {
        Object w = (World) null;
        return w; // cast w to World will be a suggested fix
    }
```

DenWav:
I don't know if there's an extension point to register custom actions for Java compilation errors or not, actually. The alternative if there wasn't would be to reimplement each the inspections for each of the cases, which isn't the best solution...If I can get an answer to my question about whether an extension point does already exists for this, then maybe.

DenWav:
The `errorQuickFixProvider` entry point seems to be totally ignored by Java, I guess.

PR #386, opened 2018-05-17
ItemStack inspections

Gegy:
This PR implements an inspection around ItemStack comparison with `ItemStack.EMPTY`. Doing this might not always give desired result, as when depleting a stack, the instance is not replaced with ItemStack.EMPTY. Instead `ItemStack.isEmpty()` should be used. This is a common mistake made when porting mods from 1.10, where stacks were nullable.

Inspection and fix visible here:
https://streamable.com/vy53o

Are there any other ItemStack related inspections that would fit within the scope of this PR?

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mcp/inspections/StackEmptyInspection.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mcp/inspections/StackEmptyInspection.kt
@@ -8,7 +8,7 @@
  * MIT License
  */
 
-package com.demonwav.mcdev.platform.forge.inspections
+package com.demonwav.mcdev.platform.mcp.inspections
 
 import com.demonwav.mcdev.util.fullQualifiedName
 import com.intellij.codeInspection.ProblemDescriptor
@@ -91,8 +91,8 @@ class StackEmptyInspection : BaseInspection() {
             }
 
             private fun isExpressionEmptyConstant(expression: PsiExpression?): Boolean {
-                val reference = expression as? PsiReferenceExpression
-                val field = reference?.resolve() as? PsiField ?: return false
+                val reference = expression as? PsiReferenceExpression ?: return false
+                val field = reference.resolve() as? PsiField ?: return false
                 return field.name == EMPTY_NAME && field.containingClass?.fullQualifiedName == STACK_FQ_NAME
             }
         }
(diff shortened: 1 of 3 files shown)


kashike:
What prevents this from being an MCP inspection?

gabizou:
This should be an MCP inspection since this can affect anything that uses MCP/Minecraft at all.

Gegy:
Moved to use an MCP inspection -- this definitely makes more sense, thanks.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/inspections/StackEmptyInspection.kt
```diff
@@ -0,0 +1,100 @@
...
+import com.intellij.psi.*
+import com.siyeh.ig.BaseInspection
+import com.siyeh.ig.BaseInspectionVisitor
+import com.siyeh.ig.InspectionGadgetsFix
+import org.jetbrains.annotations.Nls
+
+class StackEmptyInspection : BaseInspection() {
+    companion object {
+        const val STACK_FQ_NAME = "net.minecraft.item.ItemStack"
+        const val EMPTY_NAME = "EMPTY"
```
marvin-roesch: It's highly unlikely that this MCP name is going to change, but we might want to cover it using SRG names (which are also subject to change, but even less likely). See the following snippet for an example:
https://github.com/minecraft-dev/MinecraftDev/blob/a9cac682d44b8ba2ad71af4d8c69e66f637ce5b9/src/main/kotlin/com/demonwav/mcdev/i18n/translations/TranslationFunction.kt#L38-L42
I'm not sure if that's the best example, but the only one in the codebase I know off the top of my head.

marvin-roesch (changes requested):
Minor nitpick, but we should be somewhat robust against potential mapping changes.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/inspections/StackEmptyInspection.kt
Gegy: Wasn't aware classes had SRG names. Where could I check the name?
EDIT: Oh, are you just referring to the field?

DenWav: Classes do have SRG names, the SRG names just happen to match MCP names. In this case using the SRG name for the field would be more robust.

Gegy:
@PaleoCrafter Using SRG names now -- could you check this over again?

DenWav:
I'm cool with it, if @PaleoCrafter doesn't respond in a couple days it looks fine to me, I'll pull it.

marvin-roesch (approved):

marvin-roesch:
I thought I'd approved the changes already, weird. Anyways, this is fine by me, so let's get it merged :D

DenWav:
Sorry, forgetful as ever, I forgot to merge this.

PR #458, opened 2018-09-28
Turn createDependencyReducedPom off for Bukkit, Bungee & Canary Maven templates

Scarsz:
`<createDependencyReducedPom>` is an utterly useless Shade configuration option that is enabled by default. Having it on runs the risk of having IntelliJ very, very confused about the Maven projects in a given project. When it does mess IJ up, libraries aren't indexed and there are import errors everywhere. Took me a bit of debugging to figure out that that wasn't turned off by default.


DenWav:
Merged cb285bb987878b11d859796fe80f5efc003518c7

PR #459, opened 2018-10-01
[WIP] Add specialized MCP color support

marvin-roesch:
This adds support for various methods/functions in MCP code bases that take colors as parameters. Currently supported are hex color codes (encoded as integers) as well as [0,1] and [0,255] RGB vectors, all of them with and without alpha values.

Furthermore, this adds a basic system for loading configurations based on a project's underlying MC version. Currently, it's hardcoded to the color support, but I'll make it more useful as a general purpose feature. It's also a little "dumb" in that there's no concept of inheritance. Instead, configs for all versions upto the requested one are considered. This will also be addressed.

Part of the motivation behind the configuration is easier customization for the user. By exposing these configs (possibly through some auto-generated GUI), users may add support for their own methods (or those we missed).

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mcp/color/inspections/ColorComponentOutOfRangeInspection.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mcp/color/inspections/ColorComponentOutOfRangeInspection.kt
@@ -3,7 +3,7 @@
  *
  * https://minecraftdev.org
  *
- * Copyright (c) 2017 minecraft-dev
+ * Copyright (c) 2018 minecraft-dev
  *
  * MIT License
  */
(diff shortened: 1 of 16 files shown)


src/main/kotlin/com/demonwav/mcdev/platform/mcp/color/inspections/ColorComponentOutOfRangeInspection.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.intellij.codeInspection.ProblemDescriptor
+import com.intellij.openapi.module.ModuleUtilCore
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiCallExpression
+import com.siyeh.ig.BaseInspection
+import com.siyeh.ig.BaseInspectionVisitor
+import com.siyeh.ig.InspectionGadgetsFix
+import org.jetbrains.annotations.Nls
+
+class ColorComponentOutOfRangeInspection : BaseInspection() {
```
stephan-gh, line 29: Can we avoid introducing new inspections built on `BaseInspection`? In my opinion that class is pretty much obsolete and most things can be implemented with one of IntelliJ's base classes (here: `AbstractBaseJavaLocalInspectionTool`) in a nicer way.

PR #460, opened 2018-10-01
Add basic reflection references and completion

marvin-roesch:
This is a first step towards easing reflective operations using Forge utilities. It will probably have to change (for the better) when MinecraftForge/MinecraftForge#4853 gets merged.

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/forge/reflection/reference/ReflectedFieldReference.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/forge/reflection/reference/ReflectedFieldReference.kt
@@ -1,0 +1,103 @@
+/*
+ * Minecraft Dev for IntelliJ
+ *
+ * https://minecraftdev.org
+ *
+ * Copyright (c) 2017 minecraft-dev
+ *
+ * MIT License
+ */
+
+package com.demonwav.mcdev.platform.forge.reflection.reference
+
+import com.demonwav.mcdev.facet.MinecraftFacet
+import com.demonwav.mcdev.platform.mcp.McpModuleType
+import com.demonwav.mcdev.util.constantStringValue
+import com.demonwav.mcdev.util.findModule
+import com.demonwav.mcdev.util.simpleQualifiedMemberReference
+import com.intellij.codeInsight.completion.JavaLookupElementBuilder
+import com.intellij.codeInsight.hint.HintManager
+import com.intellij.lang.jvm.JvmModifier
+import com.intellij.openapi.application.runWriteAction
+import com.intellij.openapi.command.CommandProcessor
+import com.intellij.psi.JavaPsiFacade
+import com.intellij.psi.PsiClass
+import com.intellij.psi.PsiClassObjectAccessExpression
+import com.intellij.psi.PsiClassType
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/demonwav/mcdev/platform/forge/reflection/reference/ReflectedFieldReference.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    class Reference(element: PsiLiteral) : PsiReferenceBase.Poly<PsiLiteral>(element) {
+        val fieldName
+            get() = element.constantStringValue ?: ""
+
+        override fun getVariants(): Array<Any> {
+            val typeClass = findReferencedClass() ?: return arrayOf()
+
+            return typeClass.allFields.filter { !it.hasModifier(JvmModifier.PUBLIC) }
```
DenWav: Sorry, don't have my IDE open right now. What does `allFields` return? If it's anything other than a `Stream` or `Sequence`, you need to add `.asSequence()` after `allFields`. Otherwise every operation creates a new collection.

DenWav (changes requested):

src/main/kotlin/com/demonwav/mcdev/platform/forge/reflection/reference/ReflectedFieldReference.kt
```diff
@@ -0,0 +1,103 @@
...
+                }
+                .toTypedArray()
+        }
+
+        override fun multiResolve(incompleteCode: Boolean): Array<ResolveResult> {
+            val typeClass = findReferencedClass() ?: return arrayOf()
+
+            val srgManager = element.findModule()?.let { MinecraftFacet.getInstance(it) }?.getModuleOfType(McpModuleType)?.srgManager
+            val srgMap = srgManager?.srgMapNow
+            val mcpName = srgMap?.mapSrgName(fieldName)
```
stephan-gh: I'd avoid evaluating `fieldName` twice. Also the branches of the if statement below are really similar, so you can merge them with an additional variable.

Something like:

```kt
val name = fieldName
val mcpName = srgMap?.mapSrgName(name) ?: name
return typeClass.allFields.filter { it.name == name }.map(::PsiElementResolveResult).toTypedArray()
```

src/main/kotlin/com/demonwav/mcdev/platform/forge/reflection/reference/ReflectedMethodReference.kt
```diff
@@ -0,0 +1,141 @@
...
+                }
+                .toTypedArray()
+        }
+
+        override fun multiResolve(incompleteCode: Boolean): Array<ResolveResult> {
+            val typeClass = findReferencedClass() ?: return arrayOf()
+
+            val srgManager = element.findModule()?.let { MinecraftFacet.getInstance(it) }?.getModuleOfType(McpModuleType)?.srgManager
+            val srgMap = srgManager?.srgMapNow
+            val mcpName = srgMap?.mapSrgName(methodName)
```
stephan-gh: same

stephan-gh (changes requested):

DenWav:
Merged.

PR #586, opened 2019-03-23
Update Paper URL

jupjohn:
Old URL redirects to new URL so I removed the middleman in case that domain disappears.


DenWav:
Thanks

PR #613, opened 2019-05-18
Rewrite translation handling

marvin-roesch:
This is a complete rewrite of the translation (formerly i18n) component of the plugin. This was mostly motivated by the switch from `lang` to `json` files for translations and aims to make the plugin relatively agnostic with regards to translation format.

Key points of the new/improved implementation are as follows:
 * Translation files are indexed based on their locale. This allows very quick lookups and reverse searches ("where to find the translation for this key")
   * Index entries are collected both from JSON and old lang files
 * Only direct calls to (known) translation functions/constructors are respected
 * Intentions, actions etc. are agnostic of the underlying translation format

A side effect of the changes also is that sorting and "adding from default" are way more efficient than before.

The next step will be to add the option to configure what is considered a translation function. I'm mostly held up on the UI here. The configuration itself will be based on a versioned config system I've been developing.

I will also write an action for converting old lang files to the new JSON format.

Those two things will come in separate pull requests, since they're not fundamental to the underlying workings of the system.


src/main/kotlin/com/demonwav/mcdev/translations/TranslationFiles.kt
```diff
@@ -0,0 +1,299 @@
...
+import com.intellij.psi.search.GlobalSearchScope
+import com.intellij.util.DocumentUtil
+import com.intellij.util.indexing.FileBasedIndex
+import java.util.Locale
+
+object TranslationFiles {
+    private val MC_1_12_2 = SemanticVersion.release(1, 12, 2)
+
+    fun isTranslationFile(file: VirtualFile) =
+        file.mcPath?.startsWith("lang/") == true && file.fileType in listOf(LangFileType, JsonFileType.INSTANCE)
```
stephan-gh: Why did you add `== true` explicitly here?

marvin-roesch: Since `mcPath` is nullable, you need to check for an actual value.

DenWav:
Oh boy

src/main/kotlin/com/demonwav/mcdev/translations/reference/contributors.kt
```diff
@@ -58,14 +58,10 @@ class JsonReferenceContributor : PsiReferenceContributor() {
...
                     val entry = element as? JsonProperty ?: return arrayOf()
-                    val range = entry.nameElement.textRangeInParent
-                    if (range.endOffset - 3 < range.startOffset) {
-                        return emptyArray()
-                    }
                     return arrayOf(
                         TranslationReference(
                             element,
-                            element.textRange,
+                            element.nameElement.textRangeInParent.shiftRight(1).grown(-2),
```
DenWav, line 74: Just glanced through this, not doing a full review yet.

But this took me a while to figure out what was going on, could you add a clarification comment here?

DenWav (changes requested):

PR #640, opened 2019-07-29
Improve/fix Gradle Sponge project setup

RedNesto:
Wasn't sure where to do that and had to add a new parameter, sorry. 

If you have a cleaner way to do that tell me (or commit directly to the branch, I don't care).


DenWav:
Came up with a bit of a cleaner solution, but otherwise merged aa985c8446f8bc862857d6f76aae4f9b259a1da8

PR #641, opened 2019-07-30
Add Sponge plugin class inspections

RedNesto:
This PR adds two inspections to plugin classes:
- a generic inspection that covers
  - invalid plugin IDs
  - missing or private no-arg constructor (except if another constructor is `@Inject`ed)

- an `@Inject` inspection for fields, constructors and setters args:
  - checks if only one constructor is injected
  - checks if injected constructors are not optional
  - checks if type is injectable.
    - The injectable types list is not exhaustive and not configurable, but I think it covers what people use. Please note I only tested injection inspection with SpongeVanilla, SpongeForge should work too. I haven't tested with Lantern.
  - is aware of `@ConfigDir` and `@DefaultConfig`. If the injected type is `File`, `Path` or `ConfigurationLoader<CommentedConfigurationNode>` it checks one of the two annotations is present.
  - handles `Asset` injection. It will check if `@AssetId` is present.
    - it also checks if the asset exists and is not a directory.

Member annotated with `@Inject` are marked as implicitly written for fields and used for ctors and setters. Injected fields with non-optional `@Inject` are also marked as not null.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -105,6 +105,7 @@ dependencies {
     testLibs("org.jetbrains.idea:mockJDK:1.7-4d76c50")
     testLibs("org.spongepowered:mixin:0.7-SNAPSHOT:thin")
     testLibs("com.demonwav.mcdev:all-types-nbt:1.0@nbt")
+    testLibs("org.spongepowered:spongeapi:7.0.0:shaded")
 
     // For non-SNAPSHOT versions (unless Jetbrains fixes this...) find the version with:
     // println(intellij.ideaDependency.buildNumber.substring(intellij.type.length + 1))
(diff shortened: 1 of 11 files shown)


DenWav:
Could you please cleanup / squash your commits down to better commits? For example `Make Ktlint happy` and `Add license headers` aren't great commits.

You don't have to squash it all down to one commit if you don't want to, but squashing it down to a few real commits that are actually relevant to the change set rather than random noise would be really great.

RedNesto:
I'd like to add an inspection configuration for injectable types first.

About squashing commits I am not sure if I will be able to without breaking the branch, I will try anyway later, can't do it right now.

DenWav:
If you're not comfortable squashing commits that's fine, I'll handle it on merge. I'll mark this as WIP until you can add the inspection configuration.

DenWav:
Is this PR now ready for review / merge?

RedNesto:
Yup, it's ready, I don't have anything to add to it anymore.

DenWav:
Merged into dev.

PR #488, opened 2019-02-09
Add Sponge Getter support

RedNesto:
This PR adds the following features (requested in the two first bullets of #329):
- annotation's target method completion
- inspection ensuring parameter type matches targeted type (includes Optional<> handling)
- (bonus) suppression of "Optional should not be used as parameter"

Not sure if some parts of the code are really clear, I'd like to know if everything is understandable enough.

PR #652, opened 2019-09-01
Sponge Getter support fixes and improvements

RedNesto:
PR #488 added Getter filters support, but it turns out I did not tested it properly and missed many cases in which a parameter type can or cannot be used. Custom completion was not really good too.

Those issues are addressed in this PR, in addition to other tweaks and a cleanup of my old code.

_I made a better description in my first commit so please read it_


DenWav:
Looks good to me, not a big fan of the 5th commit by itself but one commit won't kill me

DenWav:
Merged
f647a5f98d172e564b0cb86aa967201fe82b04e9...508d487cc8bf2111a7b73d1445af05f78d895fce

PR #668, opened 2019-12-10
Accessor/Invoker Insight/Target support

Zidane:
Be kind, Kyle.

Signed-off-by: Chris Sanders <zidane@spongepowered.org>

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mixin/insight/AccessorLineMarkerProvider.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mixin/insight/AccessorLineMarkerProvider.kt
@@ -1,0 +1,94 @@
+/*
+ * Minecraft Dev for IntelliJ
+ *
+ * https://minecraftdev.org
+ *
+ * Copyright (c) 2019 minecraft-dev
+ *
+ * MIT License
+ */
+
+package com.demonwav.mcdev.platform.mixin.insight
+
+import com.demonwav.mcdev.asset.MixinAssets
+import com.demonwav.mcdev.platform.mixin.util.findAccessorTarget
+import com.intellij.codeInsight.daemon.GutterIconNavigationHandler
+import com.intellij.codeInsight.daemon.LineMarkerInfo
+import com.intellij.codeInsight.daemon.LineMarkerProviderDescriptor
+import com.intellij.codeInsight.daemon.MergeableLineMarkerInfo
+import com.intellij.codeInsight.daemon.impl.PsiElementListNavigator
+import com.intellij.ide.util.DefaultPsiElementCellRenderer
+import com.intellij.ide.util.MethodCellRenderer
+import com.intellij.openapi.editor.markup.GutterIconRenderer
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiField
+import com.intellij.psi.PsiIdentifier
+import com.intellij.psi.PsiMember
+import com.intellij.psi.PsiMethod
+import com.intellij.util.FunctionUtil
+import java.awt.event.MouseEvent
+import javax.swing.Icon
+
+class AccessorLineMarkerProvider : LineMarkerProviderDescriptor(), GutterIconNavigationHandler<PsiIdentifier> {
+
+    companion object {
+        private val ICON: Icon = MixinAssets.SHADOW
+        private val TOOLTIP_FUNCTION = FunctionUtil.constant<Any, String>("Go to Accessor element")
+    }
+
+    override fun getName() = "@Accessor line marker"
+    override fun getIcon() = ICON
+
+    override fun getLineMarkerInfo(element: PsiElement): LineMarkerInfo<PsiIdentifier>? {
+        if (element !is PsiMember) {
+            return null
+        }
+
+        val identifier = when (element) {
+            is PsiMethod -> element.nameIdentifier
+            is PsiField -> element.nameIdentifier
+            else -> null
+        } ?: return null
+
+        // Check if @Accessor actually has a target
+        element.findAccessorTarget() ?: return null
+        return LineMarker(identifier, this)
+    }
+
+    override fun collectSlowLineMarkers(elements: List<PsiElement>, result: Collection<LineMarkerInfo<PsiElement>>) {
+    }
+
+    override fun navigate(e: MouseEvent, elt: PsiIdentifier) {
+        val member = elt.parent as? PsiMember ?: return
+        val target = member.findAccessorTarget() ?: return
+
+        // Create correct renderer for element type
+        val renderer = when (member) {
+            is PsiMethod -> MethodCellRenderer(true)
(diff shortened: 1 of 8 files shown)


Zidane:
Note that this hasn't been tested with Invokers that trigger creating new constructor objects. Those are special and I'll see about adding it too.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/insight/AccessorLineMarkerProvider.kt
```diff
@@ -0,0 +1,94 @@
...
+        // Create correct renderer for element type
+        val renderer = when (member) {
+            is PsiMethod -> MethodCellRenderer(true)
+            is PsiField -> DefaultPsiElementCellRenderer()
+            else -> return
+        }
+
+        PsiElementListNavigator.openTargets(
+            e, arrayOf(target),
+            "Choose target class of ${member.name!!}", null, renderer
```
natanfudge, line 74: how do you know member.name != null here?

src/main/kotlin/com/demonwav/mcdev/platform/mixin/insight/InvokerLineMarkerProvider.kt
```diff
@@ -0,0 +1,94 @@
...
+        element.findInvokerTarget() ?: return null
+        return LineMarker(identifier, this)
+    }
+
+    override fun collectSlowLineMarkers(elements: List<PsiElement>, result: Collection<LineMarkerInfo<PsiElement>>) {
+    }
+
+    override fun navigate(e: MouseEvent, elt: PsiIdentifier) {
+        val member = elt.parent as? PsiMember ?: return
+        val target = member.findInvokerTarget() ?: return
```
natanfudge, line 63: The code in `getLineMarkerInfo` and `navigate` is almost identical in the accessor and the invoker.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/inspection/invoker/InvokerTargetInspection.kt
```diff
@@ -0,0 +1,55 @@
...
+            if (annotation.qualifiedName != MixinConstants.Annotations.INVOKER) {
+                return
+            }
+
+            val modifierList = annotation.owner as? PsiModifierList ?: return
+            val member = modifierList.parent as? PsiMember ?: return
+            val psiClass = member.containingClass ?: return
+            val targetClasses = psiClass.mixinTargets.ifEmpty { return }
+
+            if (resolveInvokerTarget(annotation, targetClasses, member) == null) {
```
natanfudge, line 45: Again, this class is almost exactly the same as the accessor one.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/util/Invoker.kt
```diff
@@ -0,0 +1,43 @@
...
+}
+
+@Contract(pure = true)
+fun resolveInvokerTarget(
+        invoker: PsiAnnotation,
+        targetClasses: Collection<PsiClass>,
+        member: PsiMember
+): PsiMember? {
+    val value = invoker.findDeclaredAttributeValue("value")?.constantStringValue ?: return null
+    return when (member) {
```
natanfudge, line 39: This can be an if-else expression.

natanfudge (changes requested):

src/main/kotlin/com/demonwav/mcdev/platform/mixin/insight/InvokerLineMarkerProvider.kt
Zidane: I keep it separated as they are two different annotations. We could absolutely share the code between the two but that is up to @DemonWav .

src/main/kotlin/com/demonwav/mcdev/platform/mixin/insight/AccessorLineMarkerProvider.kt
Zidane: This was borrowed from the shadow line marker code so if that is incorrect, this is too.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/util/Invoker.kt
Zidane: This is my first foray into kotlin so I made use of what I saw elsewhere in the project. This sounds more like a style issue unless the `return when` construct is inherently less efficient vs a typical if/else.

natanfudge: Well, generally the point of a when expression is for cases with 3 or more branches.

Zidane:
@DemonWav Anything holding this up from merge? Or what needs to be changed?

Earthcomputer:
Note this doesn't handle the case where the accessor/invoker target name is inferred from the accessor name. [Here](https://github.com/Earthcomputer/MinecraftDev/commit/f361611b7ba1963d0a7d9eb8f9a1978602e48e51)'s the fix for that.

DenWav:
I'm still not done with my move but things have settled down. Sorry for the ridiculous wait @Zidane.

Could you give me a simple ELI5 of this Mixin feature? What is it, quick example  in Sponge might be nice. Just so I have context of what I'm looking at when I review this. I've been out of the game for a bit so I need to get warmed up again, so to speak.

Earthcomputer:
@DemonWav [here](https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/mixin/core/tileentity/TileEntityAccessor.java) is an example accessor mixin with an accessor and an invoker in SpongeCommon. An accessor is a getter/setter of a field and an invoker delegates to the target method. The bodies or an accessor/invoker are generated by Mixin. Note also the following:
- If the target field is final, a setter accessor must also be annotated with `@Mutable` (this is a recent mixin feature added after this PR was created).
- Mixin can infer the target's name from the accessor's name, which this PR does not account of. If the accessor's name is `getXXX` or `setXXX`, then unless otherwise specified the target's name is inferred to be `XXX` (decapitalized if the entire name is not caps).
- Similarly, Mixin can infer the target's name from an invoker's name, if the invoker's name is of the form `callXXX` or `invokeXXX`.
- There are also invokers for constructors, which this PR does not handle. A constructor invoker has the return type of the target class, and the parameters of the target constructor. `@Invoker("<init>")` will target a constructor.
- Mixin can also infer that an invoker is a constructor invoker if the invoker's name starts with `new` or `create`.

DenWav:
Ah yeah I remember these now. Been a while. :)

@Zidane the code looks fine, there are cleanups & fixes and whatnot that I would like, as well as unit tests, but I'll apply those myself after I merge this so as to not waste yours or anyone else's time.

PR #670, opened 2019-12-25
authors fix in mods.toml

### LINKED ISSUE #666: Exception in mods.toml
**Minecraft Development for IntelliJ plugin version**:
2019.2-1.3.1

**IntelliJ version**: 
IntelliJ IDEA 2019.2.2 (Ultimate Edition)
Build #IU-192.6603.28, built on September 6, 2019

**Operating System**: 
Windows 10

**Description**:
I set the author and it was generated as:
```authors=""MelanX""```
There are 2 ```"``` too much. Because of that the client will not be running.

MelanX:
closes #666

### DIFF
--- a/src/main/resources/fileTemplates/j2ee/mods.toml.ft
+++ b/src/main/resources/fileTemplates/j2ee/mods.toml.ft
@@ -29,7 +29,7 @@ updateJSONURL="${UPDATE_URL}" #optional
 #credits="Thanks for this example mod goes to Java" #optional
 # A text field displayed in the mod UI
 #if (${HAS_AUTHOR_LIST})
-authors="${AUTHOR_LIST}" #optional
+authors=${AUTHOR_LIST} #optional
 #else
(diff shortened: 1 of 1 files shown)


RedNesto (changes requested):
This is not a proper fix, it only covers the case where only one author is specified. Using a comma-separated list of authors in the wizard will still generate an invalid TOML value.

The issue lies in the variable value itself, which is formatted to be a JSON array content.

The real solution would be to generate a different value based on the forge version used.
See the [MinecraftForge's mods.toml](https://github.com/MinecraftForge/MinecraftForge/blob/509a28efd3826892ba0a6bc88f0d736a4461e2bf/src/main/resources/META-INF/mods.toml#L13) for an example of what should be generated.

DenWav:
Thanks for the contribution, but @RedNesto is right. I sincerely apologize for my absence and slow response.

PR #672, opened 2020-01-02
Make static invokers not cause errors

Prospector:


Prospector:
That failed check isn't related to me right? seems the latest commit to dev has the same issue

DenWav:
Meant to merge it, oops.

PR #676, opened 2020-01-27
Partial update to 2020.1

RedNesto:
My attempt at updating the plugin for IJ 2020.1

I also fixed all deprecation warnings:
- the two components the plugin was using are removed, listeners are directly registered in the plugin descriptor. Components are deprecated and this is a step towards making the plugin dynamic
- a few deprecated usages of ToolWindowManager are replaced
- the old annotations API is deprecated (but not scheduled for removal, don't know if this is intendend), so I replaced all its usages by the new builder-style API.

The BuildSearchableOptionsTask are still failing for me so I disabled them for now (I don't even know if they are useful).

The `repository-search` plugin is an internal plugin that maven hard-depends on now, tests are failing if we do not explicitly add it in the build script.


DenWav:
Merged in branch [2020.1](https://github.com/minecraft-dev/MinecraftDev/tree/2020.1).

PR #685, opened 2020-03-16
Support several target methods in injections

### LINKED ISSUE #387: MinecraftDev does not allow method lists in annotations, but Mixin does
Using `method = {"methodOne", "methodTwo"}` in a Mixin annotation (`Inject`, `Redirect`, etc.) is supported by Mixin, but MinecraftDev makes an "Invalid member reference" error, and does not do autocompletion/ctrl-click when `method` is a list.

Version: 2018.1-1.2.5

RedNesto:
I didn't tested those changes extensively and I am not really familiar with the mixins support yet, so I hope I covered everything without breaking anything.

Fixes #387

I also fixed an NPE that was occurring in a mixin of SpongeCommon I did some tests in.


DenWav:
Any chance you could add some unit tests around this as well?

RedNesto:
Right, forgot to write some, somehow.

DenWav:
Works well in my tests (this will make a lot of people happy I think). Still need some unit tests around multiple targets, but I'll add those in a commit later.

PR #688, opened 2020-03-16
Fix mouse ungrabbing for recent game versions

### LINKED ISSUE #686: [Linux] Minecraft doesn't release mouse cursor at a breakpoint
Please include the following information in all bug reports:

* Minecraft Development for IntelliJ plugin version: 
2019.3-1.3.3

* IntelliJ version
```
IntelliJ IDEA 2019.3.3 (Ultimate Edition)
Build #IU-193.6494.35, built on February 11, 2020
Licensed to 
Subscription is active until January 22, 2021
For educational use only.
Runtime version: 11.0.5+10-b520.38 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Linux 5.3.0-7629-generic
GC: ParNew, ConcurrentMarkSweep
Memory: 1925M
Cores: 12
Registry: 
Non-Bundled Plugins: com.demonwav.minecraft-dev
```
* Operating System (this has come up enough times to be important)
Linux (Pop!_OS with GNOME desktop environment)

* Target platforms
Minecraft Forge (1.15.2)

**I saw a commit that addressed this issue (#132), but my mouse still isn't being released when the debugger reaches a breakpoint, thus making debugging impossible because of the missing cursor.**

RedNesto:
Fixes #686 

IntelliJ should take the focus anyway when breakpoints are hit (this is configurable), but it seems it does not always work.

Note that my changes ungrab the mouse correctly, but the cursor is still hidden (at least for me, on Windows 10). The method used should handle that but it doesn't work for some reason.

### DIFF
--- a/src/main/kotlin/com/demonwav/mcdev/platform/mcp/debug/UngrabMouseDebugSessionListener.kt
+++ b/src/main/kotlin/com/demonwav/mcdev/platform/mcp/debug/UngrabMouseDebugSessionListener.kt
@@ -16,10 +16,11 @@ import com.intellij.debugger.jdi.VirtualMachineProxyImpl
 import com.intellij.xdebugger.XDebugSessionListener
 import com.sun.jdi.BooleanValue
 import com.sun.jdi.ClassType
+import com.sun.jdi.ObjectReference
 
 class UngrabMouseDebugSessionListener(private val process: DebugProcessImpl) : XDebugSessionListener {
 
-    private fun grabMouse(grab: Boolean) {
+    private fun ungrabMouse() {
         val suspendContextImpl = process.debuggerContext.suspendContext ?: return
         if (suspendContextImpl.thread?.isAtBreakpoint != true) {
             return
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/demonwav/mcdev/platform/mcp/debug/UngrabMouseDebugSessionListener.kt
```diff
@@ -30,18 +31,58 @@ class UngrabMouseDebugSessionListener(private val process: DebugProcessImpl) : X
...
+            minecraftGetter,
+            emptyList()
+        ) as? ObjectReference ?: return
+
+        val mouseHelperField = minecraftClass.fieldByName("mouseHelper") ?: return
+        val mouseHelper = minecraft.getValue(mouseHelperField) as? ObjectReference ?: return
+
+        val mouseHelperClass =
+            virtualMachine.classesByName("net.minecraft.client.MouseHelper")?.singleOrNull() as? ClassType ?: return
+        val grabMouse = mouseHelperClass.methodsByName("ungrabMouse", "()V")?.singleOrNull() ?: return
```
Niko-sk2x: Maybe it would be better to not hardcode mcp method names, and try to map srg names based on current mappings if possible? (and fall back to hardcoded name if mappings not found)

Niko-sk2x:
> IntelliJ should take the focus anyway when breakpoints are hit (this is configurable), but it seems it does not always work.

Window focus is independent of mouse being grabbed on linux (or at least on some distributions). When mouse is grabbed, you can do whatever you want, if the application doesn't release it, the mouse is inaccessible until the application is killed. The only thing that works is either creating second cursor with a second mouse (it's doable, but breaks some applications that only expect one cursor) or invoking some xdotool commands that I can't remember. Intellij taking focus does nothing useful.

> Note that my changes ungrab the mouse correctly, but the cursor is still hidden (at least for me, on Windows 10). The method used should handle that but it doesn't work for some reason.

This is not even an issue on windows, as on windows doing as much as alt+tab should let the cursor free.

src/main/kotlin/com/demonwav/mcdev/platform/mcp/debug/UngrabMouseDebugSessionListener.kt
RedNesto: I thought about it, I don't know how to get the relevant module though.

DenWav: There are 2 ways to do it that I can see, the easy way, and the way I think you should do it:

_The easy way_: You can get the current `Project` from the `process` field in `UngrabMouseDebugSessionListener` and use that to find _some_ `SrgManager` in the project:

```
SrgManager.findAnyInstance(process.project)
```

 I added that method specifically for cases like this where we don't have a `Module` to go off of....except

----

_The way I think you should do it_: We do know the module we're working with, we just need to move the code around a bit.

`McpRunConfigurationExtension` is what creates the `UngrabMouseDebugSessionListener` instance and registers it as a listener, and it _does_ know which `Module` it's working with. That class extends both `ModuleDebugRunConfigurationExtension` and `DebugProcessListener`, the first does know about `Module`s, the second does not.

So to make this work those 2 classes will need to be split up as separate implementations and [here](https://github.com/minecraft-dev/MinecraftDev/blob/9fbc8c8d9d4cec0126ede2d2f2b37ed4b6653907/src/main/kotlin/com/demonwav/mcdev/platform/mcp/debug/McpRunConfigurationExtension.kt#L27) instead of passing in `this` it will need to pass in a new object that implements `DebugProcessListener` and contains a field referencing that `Module` so it can pass that `Module` to `UngrabMouseDebugSessionListener` in `processAttached`.

With that done you can get the `SrgManager` with:

```
MinecraftFacet.getInstance(module, McpModuleType)?.srgManager
```

But also keep the MCP names in as a fallback in case a SRG mapping can't be found (happens frequently enough considering how finicky FG is).

src/main/kotlin/com/demonwav/mcdev/platform/mcp/debug/UngrabMouseDebugSessionListener.kt
```diff
@@ -30,18 +35,68 @@ class UngrabMouseDebugSessionListener(private val process: DebugProcessImpl) : X
...
+            minecraftClass,
+            minecraftGetter,
+            emptyList()
+        ) as? ObjectReference ?: return
+
+        val mouseHelperField = minecraftClass.fieldByName(mcp("field_71417_B", "mouseHelper")) ?: return
+        val mouseHelper = minecraft.getValue(mouseHelperField) as? ObjectReference ?: return
+
+        val mouseHelperClass =
+            virtualMachine.classesByName("net.minecraft.client.MouseHelper")?.singleOrNull() as? ClassType ?: return
```
DenWav, line 92: Is this necessary? I think you can just do `mouseHelper.referenceType()` instead. Will help to get rid of another hardcoded string (though this one is stable so shouldn't be an issue).

DenWav (changes requested):

DenWav:
Merged in 840fde8752b7d6d8e3e69ec8d90211ee6fcc5797.

PR #691, opened 2020-03-18
Improve mixin config detection

Earthcomputer:
Many people in the Fabric community use the convention of `modid.mixins.json` rather than `mixins.modid.json`. After some discussion on the jetbrains slack, we concluded that this was the best solution.


DenWav:
I think I prefer keeping the `.json` suffix, unless there's a platform out there that prefers something else, in which case I'd rather be more specific.

If we did `(^|\.)mixins?(\.[^\.]+)*\.json$` for the regex instead it would allow multiple words / blocks and still be more flexible with order, but still require `.json` on the end. I tested it with:

```
modid.mixin.json
modid.mixins.json
mixin.modid.json
mixins.modid.json
mixin.group.modid.json
mixins.group.modid.json
group.modid.mixin.json
group.modid.mixins.json
```

And it matched all of them, but it does not match, for example, any of these:

```
modid.mixin.xml
modid.mixins.xml
mixin.modid.xml
mixins.modid.xml
mixin.group.modid.xml
mixins.group.modid.xml
group.modid.mixin.xml
group.modid.mixins.xml
```

Earthcomputer:
Whoops, that was actually a mistake to exclude the json suffix

PR #692, opened 2020-03-19
Unused mixin inspection

Earthcomputer:
This adds an inspection which highlights when a mixin is not in any mixin config file, and includes a quick fix to add the mixin to the mixin config with the longest matching package name.

The `MixinConfig` class looks a bit excessive, but it's also used by some other features in my fork, which will come in later PRs.


DenWav:
I updated the copyright year to 2020 on `dev` so a rebase should fix the CI builds.

DenWav:
I'll trust you with the `MixinConfig`  changes - I can see how it will be useful for your later changes. One style thing I noticed was a lot of places without braces on if statements, which is still an experimental ktlint rule. I'll enable the rule after I merge this and fix that myself.

The rest of the PR looks good.

PR #696, opened 2020-03-21
Fix mixin config value inspection for boxed types

Earthcomputer:
I was always getting an error on `"required": true` in the mixin config, because Mixin changed the type from `boolean` to `Boolean`.


DenWav:
Merged in e2d67bb81ff3101da9cfafc125ec6d9c09be332c.

PR #700, opened 2020-03-31
Update maven plugins dependencies.

LB45440078L:
They are outdated and should be updated to the latest one.


DenWav:
This only updates one Maven template, and I'm in-process redoing all of this so it won't be used anyways. Thanks for the contribution, just not in a good spot to use it at the moment.

PR #702, opened 2020-04-08
Accessor generators

Earthcomputer:
This PR adds a "Generate Accessor/Invoker" action, which you use in the target class to add accessors/invokers to selected members to an accessor mixin.
If there are no accessor mixins existing for a target class, then a dialog is opened for the user to create a new accessor mixin. This new accessor mixin will be automatically added to the appropriate mixin config. In the case of there being multiple existing accessor mixins, then a pop-up dialog comes up allowing the user to choose which accessor mixin to generate the accessors/invokers in.

image

This accessor mixin was automatically generated:
image

Extra things to think about when merging this:
- I had trouble getting usages of `invokeDeclaredMethod` to compile, I eventually fixed it by deleting the gradle `build` directory.
- Some code may need changes when merging into the 2020.1 branch, particularly `GenerateAccessorAction.actionPerformedImpl`, which needs to be kept in sync with the overridden method, as detailed in the javadoc.

Current extra caveats (would like to solve but not urgent, can merge without):
- [ ] Doesn't declare generic types in the mixin which are present in the target class.
- [ ] Allows you to create getter accessors for public final fields (need to find a way to prevent this that's intuitive with the UI).
- [ ] Doesn't generate a mixin config file automatically (would require modloader specific behaviour which could be implemented in the future but probably not for now).


DarkiBoi:
Ngl this is useful as fuck

DenWav:
It seems to build fine for me after a `clean build`, but will need to play with it more. Will hopefully find some time to look at this later this week.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/action/GenerateAccessorHandler.kt
```diff
@@ -0,0 +1,496 @@
...
+                val factory = JavaPsiFacade.getElementFactory(project)
+                val targetAccessible = targetClass.modifierList?.hasExplicitModifier(PsiModifier.PUBLIC) == true &&
+                    targetClass.name != null
+                val annotationText = if (targetAccessible) {
+                    "@${MixinConstants.Annotations.MIXIN}(${targetClass.qualifiedName}.class)"
+                } else {
+                    "@${MixinConstants.Annotations.MIXIN}(targets=\"${targetClass.fullQualifiedName}\")"
+                }
+                val annotation = factory.createAnnotationFromText(annotationText, clazz)
+                AddAnnotationFix(MixinConstants.Annotations.MIXIN, clazz!!, annotation.parameterList.attributes)
```
DenWav, line 302: I would really prefer that `!!` isn't used. I would be okay with either a `?.let { ... }` or, since it is truly an error condition if `clazz` is `null` at this point I am also okay with
```kt
val newClazz = clazz ?: error("Failed to create new class")
```
And using `newClazz` in place of `clazz` from that point on, removing the need for the `!!` operator. It has the same effect as `!!`, but I prefer it for 2 reasons:

 1. Contains a more clear error message
 2. Does the null-check in only one spot in the code

src/main/kotlin/com/demonwav/mcdev/platform/mixin/action/GenerateAccessorHandler.kt
```diff
@@ -0,0 +1,496 @@
...
+                            mixinClass,
+                            members
+                        )
+                    } catch (e: GenerateCodeException) {
+                        val message = e.message
+                        ApplicationManager.getApplication().invokeLater(
+                            Runnable {
+                                if (!mixinEditor.isDisposed) {
+                                    mixinEditor.caretModel.moveToOffset(offset)
+                                    HintManager.getInstance().showErrorHint(editor, message!!)
```
DenWav: Similar to my other comment, I would rather avoid the `!!` operator. I think the best solution here is to simply provide an unknown alternative message:
```kt
val message = e.message ?: "Unknown error"
```

src/main/kotlin/com/demonwav/mcdev/platform/mixin/action/GenerateAccessorHandler.kt
```diff
@@ -0,0 +1,496 @@
...
+
+    private fun methodBody(isStatic: Boolean): String {
+        return if (isStatic) {
+            "{ throw new java.lang.UnsupportedOperationException(); }"
+        } else {
+            ";"
+        }
+    }
+
+    private class HeaderPanel : JComponent() {
```
DenWav, line 472: Not a requirement or anything but this could be implemented using the Kotlin UI DSL for IntelliJ - something I would like to move my UIs to if feasible, but I don't have much experience with it so we'll see.

This would look like this probably:

```kt
private class Header {
    val gettersCheckbox = JBCheckBox("Generate Getter Accessors", true)
    val settersCheckbox = JBCheckBox("Generate Setter Accessors")
    val panel: JComponent = panel {
        row { gettersCheckbox() }
        row { settersCheckbox() }
    }

    init {
        gettersCheckbox.addItemListener(swap(settersCheckbox))
        settersCheckbox.addItemListener(swap(gettersCheckbox))
    }

    private fun swap(other: JBCheckBox) = ItemListener {
        if (it.stateChange == ItemEvent.DESELECTED) {
            other.isSelected = true
        }
    }
}
```

In my test at least the result seems identical.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/action/GenerateAccessorHandler.kt
```diff
@@ -0,0 +1,496 @@
...
+        val isEnumConstant = field is PsiEnumConstant
+        return (!isPublic || isFinal) && !isEnumConstant
+    }
+
+    private fun canHaveInvoker(method: PsiMethod): Boolean {
+        return !method.modifierList.hasExplicitModifier(PsiModifier.PUBLIC)
+    }
+
+    private fun getOrCreateAccessorMixin(project: Project, targetClass: PsiClass): PsiClass? {
+        val mixins = MixinModule.getAllMixins(project, GlobalSearchScope.projectScope(project))
```
DenWav, line 223: Later it might be nice to store Mixins (and accessor Mixins) in a cache to speed this up. It's not slow when testing in SpongeForge, but for me it seemed to hang for half a second. Nothing major, just something to consider for later.

DenWav (changes requested):
A couple things I would like changed, but a couple things aren't as important, I tried to make the distinction on each comment.

My biggest hangup with merging this is yours as well, which is the maintainability of copying the target method like that. I would like to do more research and ask some Jetbrains folks to see if there's any better option. If there's no other option then this is just what we're going to have to go with.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/action/GenerateAccessorHandler.kt
Earthcomputer: It's not possible for clazz to be null at this point. It's assigned from `JavaDirectoryService.createInstance` in the try block, which is `@NotNull`. And if the exception is caught, then it returns out of the block.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/action/GenerateAccessorHandler.kt
Earthcomputer: This was automatically converted from JetBrains' Java version in the superclass, but I suppose it's possible for message to be null. I'll implement this fix

src/main/kotlin/com/demonwav/mcdev/platform/mixin/action/GenerateAccessorHandler.kt
Earthcomputer: For me it seems like there is a bit cut off the top when I do this, it's weird
image

DenWav: Oh yeah you're right, that's strange. Maybe their UI DSL isn't so easy to use then.

src/main/kotlin/com/demonwav/mcdev/platform/mixin/action/GenerateAccessorAction.kt
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Copied from [com.intellij.codeInsight.actions.CodeInsightAction.actionPerformedImpl]
+     * except that it calls the [GenerateAccessorHandler.customInvoke] method instead of the normal one
+     */
+    override fun actionPerformedImpl(project: Project, editor: Editor?) {
+        if (editor == null) {
+            return
+        }
+        val psiFile = PsiUtilBase.getPsiFileInEditor(editor, project) ?: return
+        val handler = handler as GenerateAccessorHandler
```
DenWav, line 35: I think `val handler = this.handler as GenerateAccessorHandler` would be better here.

If you get around to updating this PR you can make this change, but I'm mostly just marking this as a note for myself when I merge this.

DenWav (approved):

src/main/kotlin/com/demonwav/mcdev/platform/mixin/action/GenerateAccessorHandler.kt
DemonWav: I generally like avoiding `!!` in general when possible, but since you've correctly stated it's not possible to be `null`, I try to find another way to re-arrange the code to show that. I took another stab at that and this is what I came up with.

```kotlin
return WriteCommandAction.writeCommandAction(project)
    .withName("Generate Accessor/Invoker")
    .withGroupId("Generate Accessor/Invoker")
    .compute<PsiClass, RuntimeException> {
        IdeDocumentHistory.getInstance(project).includeCurrentPlaceAsChangePlace()
        val clazz = try {
            JavaDirectoryService.getInstance().createInterface(pkg, name)
        } catch (e: IncorrectOperationException) {
            invokeLater {
                val message = "${CodeInsightBundle.message(
                    "intention.error.cannot.create.class.message",
                    name
                )}\n${e.localizedMessage}"
                Messages.showErrorDialog(
                    project,
                    message,
                    CodeInsightBundle.message("intention.error.cannot.create.class.title")
                )
            }
            return@compute null
        }
        val factory = JavaPsiFacade.getElementFactory(project)
        val targetAccessible = targetClass.modifierList?.hasExplicitModifier(PsiModifier.PUBLIC) == true &&
            targetClass.name != null
        val annotationText = if (targetAccessible) {
            "@${MixinConstants.Annotations.MIXIN}(${targetClass.qualifiedName}.class)"
        } else {
            "@${MixinConstants.Annotations.MIXIN}(targets=\"${targetClass.fullQualifiedName}\")"
        }
        val annotation = factory.createAnnotationFromText(annotationText, clazz)
        AddAnnotationFix(MixinConstants.Annotations.MIXIN, clazz, annotation.parameterList.attributes)
            .applyFix()

        val module = clazz.findModule() ?: return@compute null
        val configToWrite = MixinModule.getBestWritableConfigForMixinClass(
            project,
            GlobalSearchScope.moduleScope(module),
            clazz.fullQualifiedName ?: ""
        )
        configToWrite?.qualifiedMixins?.add(clazz.fullQualifiedName)

        return@compute clazz
    }
```

We can use the `computer` method in combination with using the `try { ... } catch { ... }` block as an expression and I think that's a cleaner solution.

DenWav: I generally like avoiding `!!` in general when possible, but since you've correctly stated it's not possible to be `null`, I try to find another way to re-arrange the code to show that. I took another stab at that and this is what I came up with.

```kotlin
return WriteCommandAction.writeCommandAction(project)
    .withName("Generate Accessor/Invoker")
    .withGroupId("Generate Accessor/Invoker")
    .compute<PsiClass, RuntimeException> {
        IdeDocumentHistory.getInstance(project).includeCurrentPlaceAsChangePlace()
        val clazz = try {
            JavaDirectoryService.getInstance().createInterface(pkg, name)
        } catch (e: IncorrectOperationException) {
            invokeLater {
                val message = "${CodeInsightBundle.message(
                    "intention.error.cannot.create.class.message",
                    name
                )}\n${e.localizedMessage}"
                Messages.showErrorDialog(
                    project,
                    message,
                    CodeInsightBundle.message("intention.error.cannot.create.class.title")
                )
            }
            return@compute null
        }
        val factory = JavaPsiFacade.getElementFactory(project)
        val targetAccessible = targetClass.modifierList?.hasExplicitModifier(PsiModifier.PUBLIC) == true &&
            targetClass.name != null
        val annotationText = if (targetAccessible) {
            "@${MixinConstants.Annotations.MIXIN}(${targetClass.qualifiedName}.class)"
        } else {
            "@${MixinConstants.Annotations.MIXIN}(targets=\"${targetClass.fullQualifiedName}\")"
        }
        val annotation = factory.createAnnotationFromText(annotationText, clazz)
        AddAnnotationFix(MixinConstants.Annotations.MIXIN, clazz, annotation.parameterList.attributes)
            .applyFix()

        val module = clazz.findModule() ?: return@compute null
        val configToWrite = MixinModule.getBestWritableConfigForMixinClass(
            project,
            GlobalSearchScope.moduleScope(module),
            clazz.fullQualifiedName ?: ""
        )
        configToWrite?.qualifiedMixins?.add(clazz.fullQualifiedName)

        return@compute clazz
    }
```

We can use the `computer` method in combination with using the `try { ... } catch { ... }` block as an expression and I think that's a cleaner solution.

PR #709, opened 2020-05-03
Rework project creation system

DenWav:
This refactor aims to reduce the fragility of the current project
creation system, while increasing the maintainability as well. The
maintainability issue was caused by a variety of factors, but most
notably were:

 * Many unsafe non-null assertions (`!!`)
 * Tightly coupled buildsystem and project type code
 * Significant logic duplication between project types
 * Complex shared logic with confusing conditionals

This commit attempts to address these issues by building up project
creation plans as a composed collection of discrete shareable steps.
The steps allow sharing common logic between project types without
coupling them and helps to reduce code duplication. This system does
increase declaration duplication, as each project type declares a list
of steps individually for each project configuration (typically
different configurations are created for different build systems, but
that is not always the case), but I think duplicating declarations
aren't as bad as complex duplication of logic.

Since the steps for each project type are declared in a simple list the
project plan for each project configuration type is simpler to follow as
a linear execution of steps.


src/main/kotlin/com/demonwav/mcdev/platform/forge/creator/ForgeProjectCreator.kt
```diff
@@ -0,0 +1,317 @@
...
+                config.mcVersion + "\n" +
+                config.forgeVersion + "\n" +
+                task
+
+            CreatorStep.writeText(hello, fileContents)
+        }
+    }
+
+    companion object {
+        const val HELLO = ".hello_from_mcdev"
```
Earthcomputer, line 316: I'm assuming this is left over debug code - should probably be removed

src/main/kotlin/com/demonwav/mcdev/creator/ProjectCreator.kt
```diff
@@ -0,0 +1,69 @@
...
+    protected val project
+        get() = rootModule.project
+
+    abstract fun getMainClassText(packageName: String, className: String): String
+
+    fun setupMainClassStep(): BasicMainClassStep {
+        val (packageName, className) = splitPackage(config.baseOrError.mainClass)
+        val mainClassText = getMainClassText(packageName, className)
+        return BasicMainClassStep(project, buildSystem, config.baseOrError, mainClassText)
+    }
```
Earthcomputer: Fabric doesn't really have the concept of a "main mod class", the closest thing are called "entrypoints", but some mods may have 0 entrypoints and some mods may have many, some client-only, some common and some only invoked when a certain mod is loaded.
I know you don't technically have to extend `BaseProjectCreator`, but having `FabricProjectCreator` implement `ProjectCreator` and the rest of them extend from `BaseProjectCreator` feels a bit messy. Maybe this helper method could be written some other way which doesn't force all implementations to use it?

Earthcomputer (commented):
Overall I think this refactor will make implementing new platform creators significantly simpler, in that it will reduce the number of bugs in them. Just one or two things could do with tweaking to make Fabric support in particular easier to implement, but other than that I think we're good to go.

src/main/kotlin/com/demonwav/mcdev/platform/forge/creator/ForgeProjectCreator.kt
DenWav: No, this is an unholy hack because I've as of yet found no better solution.

The issue is twofold:

 1. When IntelliJ imports the Gradle project it will totally rework the project structure according to the Gradle project. Not even the `Module` objects will be valid anymore as it will create new modules
 2. There is no simple callback or event to listen for when Gradle has finished importing the project

So this is how I get around those 2 issues. IntelliJ's Gradle support does have a powerful mechanism for hooking into Gradle configurations and pulling information out - this is how I get the location of the SRG maps in FG right now. The issue is how do I know that a  Gradle import is the first complete import of a new FG project?

That's where this file comes in. This file gets written to `<project_dir>/.gradle` with some basic information regarding the project that was created. This way, when the code runs after a Gradle import I know I should set up those run configurations if that file exists, and the info in the file lets me both set up the run configurations correctly and handle false positivies.

This is a bit annoying but IntelliJ often imports the Gradle process multiple times before the project is fully formed,  even though I asked it not to. It doesn't cause any problems so I'm not worried about it, but that means I have to wait until the module that I'm expecting to exist actually exists.

You can see the code which creates that file where you commented [here](https://github.com/minecraft-dev/MinecraftDev/blob/feature/creator/src/main/kotlin/com/demonwav/mcdev/platform/forge/creator/ForgeProjectCreator.kt#L288-L311). But this is where it's actually used: [ForgeRunConfigDataService](https://github.com/minecraft-dev/MinecraftDev/blob/feature/creator/src/main/kotlin/com/demonwav/mcdev/platform/forge/gradle/ForgeRunConfigDataService.kt)

src/main/kotlin/com/demonwav/mcdev/creator/ProjectCreator.kt
DenWav: How would a good example Fabric mod be setup from the plugin? What I mean is would you not generate a main class or would you generate multiple main classes? Or would that be dependent on the user's selections?

Earthcomputer: Yeah, dependent on user selection. In my fork, the entrypoints UI looks like this:
image
There's the entrypoint category on the left (doesn't have to be unique), class name in the middle and the interface it implements on the right.

DenWav: Ah, so that actually seems similar to how LiteLoader works. For my LiteLoader integration I just create an empty entry point, but that does seem like a nicer system. Seems like the "main class" utility could be pulled out into a separate function / class in a different way. I'll play with it a bit and see what feels good.

DenWav: I made some more changes which I think should support Fabric better. I removed `mainClass` as concept for all `ProjectConfig`s so you can do your own thing more easily with Fabric. The helper method in `BaseProjectCreator` can be used to create multiple classes, or not used at all if you like.

src/main/kotlin/com/demonwav/mcdev/creator/CreatorStep.kt
```diff
@@ -0,0 +1,125 @@
...
+    private val project: Project,
+    private val buildSystem: BuildSystem,
+    private val className: String,
+    private val classText: String
+) : CreatorStep {
+    override fun runStep(indicator: ProgressIndicator) {
+        val dirs = buildSystem.dirsOrError
+
+        runWriteTask {
+            indicator.text = "Writing main class"
```
Earthcomputer:
```suggestion
            indicator.text = "Writing class: $className"
```
That's the main one (no pun intended), but there's a couple of things in this class that need updating

DenWav: Thanks, good catch

DenWav: Finally got around to addressing this, should be good now. This PR also now includes `.gitignore` files.

Earthcomputer (approved):

PR #710, opened 2020-05-11
Implemented Support for PlaceholderAPI Expansions

darbyjack:
This pull request implements support for PlaceholderAPI Expansions. The only thing I have yet to figure out is why it's not detecting properly to show the icon. Currently, it will appear as a Spigot project.

Other than that, I believe this PR is done. Open to feedback.


Earthcomputer:
You probably should have asked before doing this. DemonWav is doing a refactor to the project creator system, which means you will have to mostly rewrite this.

DenWav:
The refactor is done, and from our discussion on Discord I think adding PlaceholderAPI as an enableable feature or extension on the existing platforms is the better option. With the new system you can potentially just create a single `CreatorStep` for PlaceholderAPI. Ping me on Discord if you have any questions.

PR #711, opened 2020-05-18
Enable bungeecord event listener generation

theminecoder:
Not sure why this was never done but this enables bungeecord based modules to use event listener generation action.


DenWav:
Oh good catch. Probably a refactor mistake.

PR #772, opened 2020-08-25
Ungrab cursor when a breakpoint is reached (Fabric edition)

Earthcomputer:


DenWav (approved):
Does yarn not need the intermediary name as well? I assume it's probably good, approved.

Earthcomputer:
The only use of that would be to protect against these names being renamed. It would mean I have to write a yarn mapping resolver, which I don't think is worth the effort given I think it's unlikely these names will change by the time obfex is ready to use.

Issue #775, opened 2020-08-30
Forge Packet class generator


Tom-The-Geek:
When I was working on a project recently, I realised that I was repeating myself multiple times every time I wanted to create a new packet and I decided to create an InteliJ template for an empty packet class.

I was wondering if this would be something the MinecraftDev plugin would be interested in including, and if it is, I will be happy to make a pull request to add it. If anyone else also has any idea for other templates I would be happy to include them.

The packet template is in a gist here: https://gist.github.com/Geek202/5f8354e120b4756ae9025688a7c5be0e

DenWav:
Seems like a good feature, apologies for the inactivity, I'm not able to stay as active as I'd always like to be.

Tom-The-Geek:
No worries, I think the reason I closed this was because #779 (the PR I made with these ideas) was merged and I felt there wasn't really a need for it to remain open

DenWav:
Ah gotcha.

PR #779, opened 2020-09-13
Code generation template ideas

Tom-The-Geek:
This pull request adds the idea that I suggested in #775, as well as a few other templates for modders who are starting out.
Currently it contains templates for:
- Forge:
   - Empty block
   - Empty item
   - A simple packet class with an `encoder`, `decoder` and `consumer`
- Fabric:
   - Empty block
   - Empty item

The `New -> MinecraftDev` option is only visible if the module has the Minecraft facet and the available templates are specific based on the modding platform in use.
A GIF of the generation and an image of the new templates section are on a gist [here](https://gist.github.com/Geek202/ac9367997f977a7590b65ffbd01fb6da).

The icon I created for the `Skeleton` section of the templates page is not final and could definitely be a lot better.
Optinions, suggestions and improvements are welcome!


DenWav:
Overall seems good, I think maybe `New -> Minecraft Class` instead of `New -> MinecraftDev`? I think that might look better.

Tom-The-Geek:
I've updated the action name in the plugin.xml and in the action class, and I agree that `New -> Minecraft Class` looks better.

Earthcomputer:
I definitely like the idea of this, creating new blocks, especially in Fabric, is a little cumbersome with several steps, and it would be nice to be able to make at least simple decorative blocks mostly with templates. It would be nice to generate simple block and item jsons like this as well for Fabric, and data generators for Forge, but those can be left out for now.

DenWav:
You need to run `./gradlew format` and fix any issues it reports that it can't automatically fix.

Tom-The-Geek:
I think I ran that task before pushing the initial code, but I'll re-run the task just to be sure

Tom-The-Geek:
Just checked, and `gradlew format` reported no issues

DenWav:
GitHub actions is bad. I've moved off it because they were failing so much. Sorry about that.

PR #782, opened 2020-09-27
Add spellcheck support for lang files

SizableShrimp:
This PR adds support for spellchecking in language files. I have tested this and confirmed it to work. I'm not the best at Kotlin, but I was using IntelliJ IDEA and don't think I programmed anything too particularly badly. I also ran `gradle format` already.

Specifically, the `LangSpellcheckingStrategy` class I implemented adds custom support for comments, translations keys, and their corresponding values.
* Translation keys (and dummy keys without a value) are treated as special identifiers, which are split by `.` and then parsed using `IdentifierSplitter`. The `IdentifierSplitter` is the same splitter used on variable names, so has inbuilt support for camelCase.
* Translation values are treated as normal plain text and parsed using `PlainTextSplitter`.
* Comments are treated as normal plain text, excluding the `#` that precedes the comment.
* Equal signs, Line endings, and the overarching entry types are skipped from tokenization/spellchecking.

Please let me know what you think; I spent a good amount of time working on this side project. I enjoyed it as I got to learn a bit of Kotlin as well as a very small view of the infrastructure of an IntelliJ plugin. I am happy to make any other changes that are suggested.

### DIFF
--- a/src/main/kotlin/translations/lang/spellcheck/LangCommentTokenizer.kt
+++ b/src/main/kotlin/translations/lang/spellcheck/LangCommentTokenizer.kt
@@ -1,0 +1,29 @@
+/*
+ * Minecraft Dev for IntelliJ
+ *
+ * https://minecraftdev.org
+ *
+ * Copyright (c) 2020 minecraft-dev
+ *
+ * MIT License
+ */
+
+package com.demonwav.mcdev.translations.lang.spellcheck
+
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.impl.source.tree.LeafPsiElement
+import com.intellij.spellchecker.inspections.PlainTextSplitter
+import com.intellij.spellchecker.tokenizer.TokenConsumer
+import com.intellij.spellchecker.tokenizer.Tokenizer
+
+class LangCommentTokenizer : Tokenizer<LeafPsiElement>() {
+    override fun tokenize(element: LeafPsiElement, consumer: TokenConsumer?) {
+        val text = element.text
+        val startOffset = when {
+            text.startsWith('#') -> 1
+            else -> 0
+        }
+        val range = TextRange(startOffset, text.length)
+        consumer?.consumeToken(element, text, false, 0, range, PlainTextSplitter.getInstance())
+    }
+}
--- a/src/main/kotlin/translations/lang/spellcheck/LangCommentTokenizer.kt
+++ b/src/main/kotlin/translations/lang/spellcheck/LangCommentTokenizer.kt
@@ -17,13 +17,13 @@ import com.intellij.spellchecker.tokenizer.TokenConsumer
 import com.intellij.spellchecker.tokenizer.Tokenizer
 
 class LangCommentTokenizer : Tokenizer<LeafPsiElement>() {
-    override fun tokenize(element: LeafPsiElement, consumer: TokenConsumer?) {
+    override fun tokenize(element: LeafPsiElement, consumer: TokenConsumer) {
         val text = element.text
         val startOffset = when {
             text.startsWith('#') -> 1
             else -> 0
         }
         val range = TextRange(startOffset, text.length)
-        consumer?.consumeToken(element, text, false, 0, range, PlainTextSplitter.getInstance())
+        consumer.consumeToken(element, text, false, 0, range, PlainTextSplitter.getInstance())
     }
 }
--- a/src/main/kotlin/translations/lang/spellcheck/LangKeySplitter.kt
+++ b/src/main/kotlin/translations/lang/spellcheck/LangKeySplitter.kt
@@ -16,7 +16,7 @@ import com.intellij.spellchecker.inspections.IdentifierSplitter
 import com.intellij.util.Consumer
 
 object LangKeySplitter : BaseSplitter() {
-    override fun split(text: String?, range: TextRange, consumer: Consumer<TextRange>?) {
+    override fun split(text: String?, range: TextRange, consumer: Consumer<TextRange>) {
         if (text == null) {
             return
         }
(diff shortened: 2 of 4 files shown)


SizableShrimp:
Any update on this pull request and/or the glitchy windows JDK 11 build check?

src/main/kotlin/translations/lang/spellcheck/LangCommentTokenizer.kt
```diff
@@ -0,0 +1,29 @@
...
+package com.demonwav.mcdev.translations.lang.spellcheck
+
+import com.intellij.openapi.util.TextRange
+import com.intellij.psi.impl.source.tree.LeafPsiElement
+import com.intellij.spellchecker.inspections.PlainTextSplitter
+import com.intellij.spellchecker.tokenizer.TokenConsumer
+import com.intellij.spellchecker.tokenizer.Tokenizer
+
+class LangCommentTokenizer : Tokenizer<LeafPsiElement>() {
+    override fun tokenize(element: LeafPsiElement, consumer: TokenConsumer?) {
```
RedNesto: Looks like `consumer` is never null, you should be able to remove the nullability marker safely.

src/main/kotlin/translations/lang/spellcheck/LangKeySplitter.kt
```diff
@@ -0,0 +1,37 @@
...
+
+package com.demonwav.mcdev.translations.lang.spellcheck
+
+import com.intellij.openapi.util.TextRange
+import com.intellij.spellchecker.inspections.BaseSplitter
+import com.intellij.spellchecker.inspections.IdentifierSplitter
+import com.intellij.util.Consumer
+
+object LangKeySplitter : BaseSplitter() {
+    override fun split(text: String?, range: TextRange, consumer: Consumer<TextRange>?) {
```
RedNesto: Same here, `consumer` does not need to be nullable.

src/main/kotlin/translations/lang/spellcheck/LangSpellcheckingStrategy.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.intellij.spellchecker.tokenizer.Tokenizer
+import com.intellij.spellchecker.tokenizer.TokenizerBase
+
+class LangSpellcheckingStrategy : SpellcheckingStrategy() {
+    private val langCommentTokenizer = LangCommentTokenizer()
+    private val langKeyTokenizer = TokenizerBase<LeafPsiElement>(LangKeySplitter)
+
+    override fun isMyContext(element: PsiElement) = MCLangLanguage.`is`(element.language)
+
+    override fun getTokenizer(element: PsiElement?): Tokenizer<*> {
```
RedNesto, line 27: `element` is never null either from what I saw in the method usages.

RedNesto (approved):
Hi, thanks for the PR. Tested and it works really well 👍 

It also looks properly implemented, with good use of Kotlin's idioms.

I am not familiar with IJ's spellchecker API I can't comment on whether this could be done in a better way.

My comments are really minor, just small nitpicks.

As for the check failure I have no idea why it happened, I'll rerun the jobs and hope it works this time.

SizableShrimp:
Thanks for the review! I'm glad that I properly used Kotlin in this instance. I should really read up on it and learn more about it; it's quite powerful! I'll fix up these few issues and add another commit.

src/main/kotlin/translations/lang/spellcheck/LangCommentTokenizer.kt
SizableShrimp: I have removed this nullability marker.

src/main/kotlin/translations/lang/spellcheck/LangSpellcheckingStrategy.kt
SizableShrimp: I kept the nullability marker here because it could be possible that a null element is passed in depending on the PSI tree and file structure of the lang file, along with it being possible to change in the future. Right now, this shouldn't be possible but I still kept it.

src/main/kotlin/translations/lang/spellcheck/LangKeySplitter.kt
SizableShrimp: I have removed this nullability marker.

SizableShrimp:
Just pushed another commit after testing that it didn't crash. All checks passed, so this should be good to go.

DenWav:
The code looks good, the only question I have here is how does the spellchecker handle other languages? This is targetting something specifically which will intentionally have lots of other languages in it, but when I open a non-english file it highlights basically everything as a typo.

Is there a way to tell IntelliJ which langauge it should be based on the name of the file?

SizableShrimp:
I'm not sure; I can investigate this. I forgot to consider that these are language files after all. :P

SizableShrimp:
It seems that the only way to do this is by the user installing the language to their dictionary themselves through use of the `Hunspell` plugin, see [this](https://www.jetbrains.com/help/idea/spellchecking.html#configure-the-dictionaries-to-use). Not sure how to easily integrate this into this plugin. I haven't dabbled in any of the other code; maybe it could be possible to suggest the user install the Hunspell plugin on non-English language files with a notification bar at the top of the file? However, most of the time, translations do not come from the mod/plugin creator themselves but external creators instead. They probably don't have much use to open these files.

I haven't used the automatic extraction of a string literal into the lang files yet, but I assume the only reason to open other language files would be to re-order the translation key to where the creator wants in the file after extracting a string literal. Maybe this support could also be added to the plugin? I would personally find it useful to be able to sort my English translation keys and it apply those changes to all the files, including adding whitespace and comments to the other files. This way, creators won't have much use to open the files and therefore won't need the Hunspell plugin.

PR #787, opened 2020-10-09
[Mixin] Always enable "Fix method parameters" fix

### LINKED ISSUE #766: Mixin: add "Fix parameters" option for non-empty injection parameter lists
Currently, the "Fix parameters" option in the quick fix list for a Mixin injection appears only when the its parameter list is empty, but it would make more sense and be more convenient if it would appear when the parameter list is not empty so that the parameters do not have to be deleted every time that somebody wants to use the said option.

RedNesto:
The fix now keeps every parameters after the CallbackInfo to not remove captured locals.

I am not sure if my solution is entirely correct, let me know if there are cases where it won't work.

Fixes #766

### DIFF
--- a/src/main/kotlin/platform/mixin/inspection/injector/InvalidInjectorMethodSignatureInspection.kt
+++ b/src/main/kotlin/platform/mixin/inspection/injector/InvalidInjectorMethodSignatureInspection.kt
@@ -12,6 +12,8 @@ package com.demonwav.mcdev.platform.mixin.inspection.injector
 
 import com.demonwav.mcdev.platform.mixin.inspection.MixinInspection
 import com.demonwav.mcdev.platform.mixin.reference.MethodReference
+import com.demonwav.mcdev.platform.mixin.util.MixinConstants
+import com.demonwav.mcdev.util.fullQualifiedName
 import com.demonwav.mcdev.util.isErasureEquivalentTo
 import com.demonwav.mcdev.util.synchronize
 import com.intellij.codeInsight.intention.QuickFixFactory
@@ -21,9 +23,11 @@ import com.intellij.codeInspection.ProblemsHolder
 import com.intellij.openapi.project.Project
 import com.intellij.psi.JavaElementVisitor
 import com.intellij.psi.JavaPsiFacade
+import com.intellij.psi.PsiClassType
 import com.intellij.psi.PsiElementVisitor
 import com.intellij.psi.PsiMethod
 import com.intellij.psi.PsiModifier
+import com.intellij.psi.PsiParameter
 import com.intellij.psi.PsiParameterList
 import com.intellij.psi.codeStyle.JavaCodeStyleManager
 import com.intellij.psi.codeStyle.VariableKind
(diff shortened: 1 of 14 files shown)


src/main/kotlin/platform/mixin/inspection/injector/InvalidInjectorMethodSignatureInspection.kt
```diff
@@ -124,28 +120,44 @@ class InvalidInjectorMethodSignatureInspection : MixinInspection() {
...
-                    } else {
-                        listOf()
-                    }
+            // We want to preserve captured locals
+            val locals = parameters.parameters.dropWhile {
+                val type = it.type as? PsiClassType ?: return@dropWhile true
+                val fqname = type.fullQualifiedName ?: return@dropWhile true
+                return@dropWhile fqname != MixinConstants.Classes.CALLBACK_INFO &&
+                    fqname != MixinConstants.Classes.CALLBACK_INFO_RETURNABLE
+            }.drop(1) // the first element in the list is the CallbackInfo but we don't want it
```
DenWav, line 143: Should we check that this is in-fact the case or is this already guaranteed before we get to this point? (Thinking about incorrect code here).

As far as I can tell we actually only hit this point when the `checkParameters()` method fails, so I think a sufficieintly broken method could be missing the callback info.

I noticed in your unit tests all of the test cases do have this parameter.

src/main/kotlin/platform/mixin/inspection/injector/InvalidInjectorMethodSignatureInspection.kt
```diff
@@ -124,28 +120,44 @@ class InvalidInjectorMethodSignatureInspection : MixinInspection() {
...
+            val newParams = expected.flatMapTo(mutableListOf<PsiParameter>()) {
+                if (it.default) {
+                    it.parameters?.map {
+                        JavaPsiFacade.getElementFactory(project).createParameter(
+                            it.name ?: JavaCodeStyleManager.getInstance(project)
+                                .suggestVariableName(VariableKind.PARAMETER, null, null, it.type).names
+                                .firstOrNull() ?: "unknown",
```
DenWav: What about `var<index>` instead of `unknown`?

DenWav (changes requested):

src/main/kotlin/platform/mixin/inspection/injector/InvalidInjectorMethodSignatureInspection.kt
RedNesto: Good question, this list contains the locals, we only distinguish those thanks to the CallbackInfo so if this list has actual locals, it will also always contain the CallbackInfo.

If there's no CallbackInfo then what should be the locals are stripped as they are considered part of the mandatory parameters by the inspection.


I wanted to add more tests but I kinda (totally) forgot about it.

DenWav: That makes sense, thanks!

PR #811, opened 2020-12-30
Port some features shared by all frameworks to UAST

RedNesto:
Namely:
- color line markers & annotator
- event listener line marker & annotators
- plugin line markers

I target 2020.3 because it has the most complete UAST implementation for all major JVM language plugins; I ran into some major issues with previous versions (some even affected Java code.)

I encountered a few minor problems so far:
* Annotations & color line markers do not work in Groovy, but are fine in Java, Kotlin and Scala. I suspect this is a bug in the Groovy plugin and given the language is not really used in modding AFAIK I will ignore this problem.

* The Scala plugin doesn't provide a `UastCodeGenerationPlugin` so we can't replace color constants. Vector-based Sponge `Color`s do not work either because of an NPE in their PsiElementFactory.

### DIFF
--- a/src/main/kotlin/insight/ColorAnnotator.kt
+++ b/src/main/kotlin/insight/ColorAnnotator.kt
@@ -18,6 +18,8 @@ import com.intellij.openapi.editor.markup.TextAttributes
 import com.intellij.psi.PsiElement
 import java.awt.Color
 import java.awt.Font
+import org.jetbrains.uast.UIdentifier
+import org.jetbrains.uast.toUElementOfType
 
 class ColorAnnotator : Annotator {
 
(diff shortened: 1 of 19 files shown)


DenWav:
Those issues don't sound too major to me. Looking forward to looking at this.

DenWav (approved):
Sorry for leaving this for so long.

Just looked at this in depth and I'm really impressed, it looks great and it's still pretty clean, while greatly expanding mcdev's feature support. Very well done!

I'll leave this open just in case there's more you want to do on it - I see you mentioned this in #846, so not sure if you want to do that separately or update this PR. I'm marking this as approved, so go ahead and merge it when you're ready.

RedNesto:
Great :)

I just pushed a commit to fix a Ktlint violation.

I'll fix the other issue separately, getting this PR merged as-is is already good enough.

PR #817, opened 2021-01-18
Small addition and fix in Mixin Module

### LINKED ISSUE #805: Validate type of argument for `@ModifyArg` mixins
Should be simple, inspection + quickfix.

LlamaLad7:
Adds parameter and return type checking for @ModifyArgs
Resolves outer class constructor param being required; it is now treated as part of the method's optional parameters

### DIFF
--- a/src/test/kotlin/platform/mixin/InvalidInjectorMethodSignatureFixTest.kt
+++ b/src/test/kotlin/platform/mixin/InvalidInjectorMethodSignatureFixTest.kt
@@ -33,7 +33,7 @@ class InvalidInjectorMethodSignatureFixTest : BaseMixinTest() {
 
                     class MixedInSimple {
                         public void simpleMethod(String string, int i) {
-                            int i = Integer.parseInt("FF", 16);
+                            int testInt = Integer.parseInt("FF", 16);
                         }
                     }
                     """,
--- a/src/test/kotlin/platform/mixin/InvalidInjectorMethodSignatureFixTest.kt
+++ b/src/test/kotlin/platform/mixin/InvalidInjectorMethodSignatureFixTest.kt
@@ -33,6 +33,7 @@ class InvalidInjectorMethodSignatureFixTest : BaseMixinTest() {
 
                     class MixedInSimple {
                         public void simpleMethod(String string, int i) {
+                            int i = Integer.parseInt("FF", 16);
                         }
                     }
                     """,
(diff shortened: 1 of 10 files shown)


LlamaLad7:
This validates the type arguments of `@ModifyArgs` mixins, not `@ModifyArg` mixins. I was initially planning to do `@ModifyArg` as well, but they get special treatment in that you can either take a single arg, or all the method's args, and I couldn't find a way to represent that with the `ParameterGroup`s currently in place.

RedNesto (changes requested):
Could you add tests for ModifyArgs ?
The rest LGTM.

LlamaLad7:
In the process of writing a test I discovered that the fix was inserting the fully-qualified `Args` type instead of using the import.
This is only reproduceable inside the test environment, and when using `runIde` it all works fine.
It appears that not using the `thin` version of the Mixin dependency fixes this (perhaps the thin version doesn't include that type?)
Would you rather I allow the test to insert the fully-qualified type, or depend on the non-thin version of Mixin?
Thanks.

RedNesto:
I'd say use the non-thin version for now, but I'd like input from @DemonWav about this.

src/test/kotlin/platform/mixin/InvalidInjectorMethodSignatureFixTest.kt
```diff
@@ -33,6 +33,7 @@ class InvalidInjectorMethodSignatureFixTest : BaseMixinTest() {
 
                     class MixedInSimple {
                         public void simpleMethod(String string, int i) {
+                            int i = Integer.parseInt("FF", 16);
```
RedNesto: The variable should be called something else, `i` is already used as a parameter. I don't know if it can conflict with anything, just in case.

RedNesto (changes requested):

src/test/kotlin/platform/mixin/InvalidInjectorMethodSignatureFixTest.kt
LlamaLad7: Whoops I didn't see that.

DenWav:
That's interesting, iirc thin means it only includes the API classes, so it should be included. It's been a while though - @Mumfrey would know either way.

DenWav (approved):
I think this looks good to go.

RedNesto (approved):

Earthcomputer (approved):

PR #834, opened 2021-02-27
basic adventure color support

### LINKED ISSUE #754: Add colour support for Adventure
The Kyori adventure library is a popular text library and is used in many popular platforms (such as the new API8 of Sponge). Adding colour support for this library would be beneficial.

https://github.com/KyoriPowered/adventure/blob/master/api/src/main/java/net/kyori/adventure/text/format/TextColor.java
https://github.com/KyoriPowered/adventure/blob/master/api/src/main/java/net/kyori/adventure/text/format/NamedTextColor.java

kashike:


src/main/kotlin/platform/adventure/AdventureModule.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import com.demonwav.mcdev.facet.MinecraftFacet
+import com.demonwav.mcdev.platform.AbstractModule
+import com.demonwav.mcdev.platform.PlatformType
+import com.demonwav.mcdev.platform.mixin.MixinModuleType
+import javax.swing.Icon
+
+class AdventureModule(facet: MinecraftFacet) : AbstractModule(facet) {
+  override val moduleType = MixinModuleType
+  override val type = PlatformType.MIXIN
```
RedNesto: Are you sure about using these values ? :p

src/main/kotlin/platform/adventure/framework/AdventureLibraryKind.kt
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) 2020 minecraft-dev
+ *
+ * MIT License
+ */
+
+package com.demonwav.mcdev.platform.adventure.framework
+
+import com.intellij.openapi.roots.libraries.LibraryKind
+
+val ADVENTURE_LIBRARY_KIND: LibraryKind = LibraryKind.create("adventure-api")
```
RedNesto: This should use our `libraryKind` top-level function like:
https://github.com/minecraft-dev/MinecraftDev/blob/f30d1d798d1439da6e708e3e8d21ccda3e0659f7/src/main/kotlin/platform/bukkit/framework/BukkitLibraryKind.kt#L16

RedNesto (changes requested):
This looks pretty good.

For some reason Adventure isn't auto-detected and I don't know why.
I tried with spongeapi 8 alone and with an explicit dependency on adventure-api.

RedNesto (approved):

PR #847, opened 2021-04-11
Fold invoker and accessor Mixin method calls to appear as the target method name

DenWav:
This folder attempts to make reading code which use invokers a little
cleaner by showing the user the method they are actually calling in the
end, rather than the large cast and weird method name of the invoker.

Example:
fold

Accessor folding example:
fold2

To do:
 - [x] Implement accessor getter support (this should be easy)
 - [x] Implement accessor setter support (not sure about this one)

Invoker and accessor cast  folding is enabled by default, and invoker method calls is enabled by default. I think setter folding for accessors looks a little weird so I've left that disabled by default.


Earthcomputer:
I would certainly have this enabled

DenWav:
Should be ready to go, will probably plan to merge this on the weekend, in case anyone with any opinions comments on their  preferred default settings.

SizableShrimp:
This looks really nice! I have no objections to it being enabled by default, and I doubt anyone else would be either.

RedNesto (approved):
Looks good

kb-1000:
Does it support static accessors/invokers?

Earthcomputer:
Looks like it doesn't yet, support for them would also be nice :+1:

dexman545:
Looks very good.

LlamaLad7:
Would also be nice to have support for factory invokers.

DenWav:
Static and factory accessors/invokers must be newer features (in that I don't know about them). Will see what I can find out, as long as the accessor and invoker targets could be added to these methods:

 * https://github.com/minecraft-dev/MinecraftDev/blob/dev/src/main/kotlin/platform/mixin/util/Accessor.kt#L27-L55
 * https://github.com/minecraft-dev/MinecraftDev/blob/dev/src/main/kotlin/platform/mixin/util/Invoker.kt#L27-L55

These folds would work.

LlamaLad7:
Perhaps this would be best suited as a separate idea, but it would also be cool to have autocomplete for fields and methods we have exposed with accessors/invokers, so that using the autocomplete option would auto-cast to our accessor Mixin and call the corresponding method.

RedNesto:
@LlamaLad7 you should open a new issue ;)

LlamaLad7:
Yeah fair enough, will do.

PR #850, opened 2021-04-21
Force UTF-8 in processResource task

Toshimichi0915:
processResource task uses Windows system encoding(which isn't UTF-8 by default).
People who use other than English might get confused about why their plugins/mods not being loaded properly.
This PR forces UTF-8 in processResource task


Toshimichi0915:
closed because I realized that the template doesn't have encoding settings at all.
Maybe most users don't care about their encodings sadly

PR #903, opened 2021-05-23
Fix AT generation for members in subclasses

SizableShrimp:
Currently, ATs generated by this plugin are invalid if the field or method is inside of a subclass. This fixes the problem and correctly generates AT entries. (Tested and works)

### DIFF
--- a/src/main/kotlin/platform/mcp/actions/CopyAtAction.kt
+++ b/src/main/kotlin/platform/mcp/actions/CopyAtAction.kt
@@ -25,9 +25,11 @@ class CopyAtAction : SrgActionBase() {
     override fun withSrgTarget(parent: PsiElement, srgMap: McpSrgMap, e: AnActionEvent, data: ActionData) {
         when (parent) {
             is PsiField -> {
-                if (parent.containingClass == null)
+                if (parent.containingClass == null) {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/platform/mcp/actions/CopyAtAction.kt
```diff
@@ -25,19 +25,33 @@ class CopyAtAction : SrgActionBase() {
     override fun withSrgTarget(parent: PsiElement, srgMap: McpSrgMap, e: AnActionEvent, data: ActionData) {
         when (parent) {
             is PsiField -> {
+                if (parent.containingClass == null) {
+                    return showBalloon("No SRG name found", e)
+                }
+                val classSrg = srgMap.getSrgClass(parent.containingClass!!) ?: return showBalloon(
```
Earthcomputer: Is this `!!` necessary? Maybe you can remove it by extracting it to a local variable.

SizableShrimp: Fixed

Earthcomputer (approved):

RedNesto (approved):
Sorry for the force push, I messed up my branches a little :c

PR #928, opened 2021-05-28
Sponge 8 templates

### LINKED ISSUE #918: Add SpongeAPI 8 plugin template
SpongeAPI 8 is [planned to be frozen in July](https://forums.spongepowered.org/t/status-update-23rd-may-2021/39234).

Templates can be based on [SpongePowered/sponge-plugin-template](https://github.com/SpongePowered/sponge-plugin-template) and we can already use API snapshots with `8.0.0-SNAPSHOT`.

RedNesto:
This adds SpongeAPI 8 plugin templates to the project creator.

The templates are based on https://github.com/SpongePowered/sponge-plugin-template
They are using the Gradle Kotlin DSL and Gradle 7, so I had to adapt some Gradle steps to take the Kotlin DSL into account.
Gradle 7 also breaks some old templates (those using the `maven` plugin or `compile` configurations for example.) This is an issue with multi-project setups and will be fixed separately.

The maven templates do not generate a plugin manifest (the Sponge 7 one didn't either so I don't consider this an important issue.)

Fixes #918

### DIFF
--- a/src/main/kotlin/platform/sponge/creator/Sponge8ProjectCreator.kt
+++ b/src/main/kotlin/platform/sponge/creator/Sponge8ProjectCreator.kt
@@ -69,17 +69,25 @@ class Sponge8GradleCreator(
     }
 
     override fun getMultiModuleSteps(projectBaseDir: Path): Iterable<CreatorStep> {
-        TODO()
-        // val mainClassStep = setupMainClassStep()
-        //
-        // val buildText = SpongeTemplate.applySubBuildGradle(project, buildSystem)
-        // val files = GradleFiles(buildText, null, null)
-        //
-        // return listOf(
-        //     setupDependencyStep(),
-        //     CreateDirectoriesStep(buildSystem, rootDirectory),
-        //     GradleSetupStep(project, rootDirectory, buildSystem, files),
-        //     mainClassStep
-        // )
+        val mainClassStep = setupMainClassStep()
+
+        val buildText = Sponge8Template.applySubBuildGradle(
+            project,
+            buildSystem,
+            config.spongeApiVersion,
+            config.pluginName,
+            config.mainClass,
+            config.description,
+            config.website,
+            config.authors,
+            config.dependencies
+        )
+        val files = GradleFiles(buildText, null, null)
+
+        return listOf(
+            CreateDirectoriesStep(buildSystem, rootDirectory),
+            GradleSetupStep(project, rootDirectory, buildSystem, files, true),
+            mainClassStep
+        )
     }
 }
--- a/src/main/kotlin/platform/sponge/creator/Sponge8ProjectCreator.kt
+++ b/src/main/kotlin/platform/sponge/creator/Sponge8ProjectCreator.kt
@@ -107,13 +107,7 @@ class Sponge8GradleCreator(
         val buildText = Sponge8Template.applyBuildGradle(
             project,
             buildSystem,
-            config.spongeApiVersion,
-            config.pluginName,
-            config.mainClass,
-            config.description,
-            config.website,
-            config.authors,
-            config.dependencies
+            config
         )
         val propText = Sponge8Template.applyGradleProp(project)
         val settingsText = Sponge8Template.applySettingsGradle(project, buildSystem.artifactId)
@@ -135,13 +129,7 @@ class Sponge8GradleCreator(
         val buildText = Sponge8Template.applySubBuildGradle(
             project,
             buildSystem,
-            config.spongeApiVersion,
-            config.pluginName,
-            config.mainClass,
-            config.description,
-            config.website,
-            config.authors,
-            config.dependencies
+            config
         )
         val files = GradleFiles(buildText, null, null)
 
--- a/src/main/kotlin/platform/sponge/creator/Sponge8ProjectCreator.kt
+++ b/src/main/kotlin/platform/sponge/creator/Sponge8ProjectCreator.kt
@@ -1,0 +1,85 @@
+package com.demonwav.mcdev.platform.sponge.creator
+
+import com.demonwav.mcdev.creator.BaseProjectCreator
+import com.demonwav.mcdev.creator.BasicJavaClassStep
+import com.demonwav.mcdev.creator.CreateDirectoriesStep
+import com.demonwav.mcdev.creator.CreatorStep
+import com.demonwav.mcdev.creator.buildsystem.BuildSystem
+import com.demonwav.mcdev.creator.buildsystem.gradle.BasicGradleFinalizerStep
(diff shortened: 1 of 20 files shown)


src/main/kotlin/platform/sponge/creator/Sponge8ProjectCreator.kt
```diff
@@ -0,0 +1,154 @@
...
+        val buildText = Sponge8Template.applyBuildGradle(
+            project,
+            buildSystem,
+            config.spongeApiVersion,
+            config.pluginName,
+            config.mainClass,
+            config.description,
+            config.website,
+            config.authors,
+            config.dependencies
```
DenWav: Might be better to just pass the `config` object in instead of destructuring it here.

src/main/kotlin/platform/sponge/creator/Sponge8ProjectCreator.kt
```diff
@@ -0,0 +1,154 @@
...
+        val buildText = Sponge8Template.applySubBuildGradle(
+            project,
+            buildSystem,
+            config.spongeApiVersion,
+            config.pluginName,
+            config.mainClass,
+            config.description,
+            config.website,
+            config.authors,
+            config.dependencies
```
DenWav: Same as above.

src/main/kotlin/util/MinecraftTemplates.kt
```diff
@@ -66,6 +66,13 @@ class MinecraftTemplates : FileTemplateGroupDescriptorFactory {
             spongeGroup.addTemplate(FileTemplateDescriptor(SPONGE_SETTINGS_GRADLE_TEMPLATE))
             spongeGroup.addTemplate(FileTemplateDescriptor(SPONGE_POM_TEMPLATE))
             spongeGroup.addTemplate(FileTemplateDescriptor(SPONGE_SUBMODULE_POM_TEMPLATE))
+            spongeGroup.addTemplate(FileTemplateDescriptor(SPONGE8_MAIN_CLASS_TEMPLATE))
+            spongeGroup.addTemplate(FileTemplateDescriptor(SPONGE8_BUILD_GRADLE_TEMPLATE))
+            // spongeGroup.addTemplate(FileTemplateDescriptor(SPONGE8_SUBMODULE_BUILD_GRADLE_TEMPLATE))
+            spongeGroup.addTemplate(FileTemplateDescriptor(SPONGE8_GRADLE_PROPERTIES_TEMPLATE))
+            spongeGroup.addTemplate(FileTemplateDescriptor(SPONGE8_SETTINGS_GRADLE_TEMPLATE))
+            // spongeGroup.addTemplate(FileTemplateDescriptor(SPONGE8_POM_TEMPLATE))
+            // spongeGroup.addTemplate(FileTemplateDescriptor(SPONGE8_SUBMODULE_POM_TEMPLATE))
```
DenWav: Why are these commented out?

DenWav (approved):
Looks good.

src/main/kotlin/util/MinecraftTemplates.kt
RedNesto: I simply forgot to remove them. Thought I needed different POM templates when I started working on this, but the current ones work fine.

src/main/kotlin/platform/sponge/creator/Sponge8Template.kt
```diff
@@ -0,0 +1,93 @@
...
+    fun applyBuildGradle(
+        project: Project,
+        buildSystem: BuildSystem,
+        config: SpongeProjectConfig
+    ): String {
+        val props = mutableMapOf(
+            "GROUP_ID" to buildSystem.groupId,
+            "PLUGIN_ID" to buildSystem.artifactId,
+            "PLUGIN_VERSION" to buildSystem.version,
+            // SpongeGradle 1.1.0 adds the -SNAPSHOT suffix itself
```
zml2008, line 69: this will probably:tm: change at some point in the future (just as a heads-up at the moment)

src/main/kotlin/platform/sponge/creator/SpongeProjectConfig.kt
```diff
@@ -47,14 +49,31 @@ class SpongeProjectConfig : ProjectConfig(), MavenCreator, GradleCreator {
...
+            SpongeGradleCreator(rootDirectory, module, buildSystem, this)
+        } else {
+            Sponge8GradleCreator(rootDirectory, module, buildSystem, this)
+        }
+    }
+
+    override fun configureRootGradle(rootDirectory: Path, buildSystem: GradleBuildSystem) {
+        val apiVersion = SemanticVersion.parse(spongeApiVersion)
+        if (apiVersion >= SpongeConstants.API8) {
+            buildSystem.gradleVersion = SemanticVersion.release(7, 0, 2)
```
zml2008, line 76: what is the default Gradle version? I believe SpongeGradle should work fine with gradle 6.x if projects prefer that

zml2008 (commented):
a few notes.

For Maven, the reason the API 7 template didn't include a plugin file was because it was previously generated by annotation processor. This is no longer the case in API 8. You'll probably want to include a `META-INF/plugins.json` as part of the template for maven users.

an example (generated for the example plugin):

```json
{
    "plugins": [
        {
            "loader": "java_plain",
            "id": "example",
            "name": "Example",
            "version": "1.0-SNAPSHOT",
            "main-class": "org.spongepowered.example.Example",
            "description": "Just testing things...",
            "links": {
                "homepage": "https://spongepowered.org",
                "source": "https://spongepowered.org/source",
                "issues": "https://spongepowered.org/issues"
            },
            "contributors": [
                {
                    "name": "Spongie",
                    "description": "Lead Developer"
                }
            ],
            "dependencies": [
                {
                    "id": "spongeapi",
                    "version": "8.0.0",
                    "load-order": "AFTER",
                    "optional": false
                }
            ]
        }
    ]
}
```

src/main/kotlin/platform/sponge/creator/Sponge8Template.kt
RedNesto: I know, I'm still following pretty much everything that happens on SpongePowered repos so I should be able to update it not long after this changes ^^

src/main/kotlin/platform/sponge/creator/SpongeProjectConfig.kt
RedNesto: The creator uses 5.6.1 by default which is quite old. (I plan to update most templates and the default Gradle version later)

That and I'd rather stick with what the official template uses.

RedNesto:
Wow I somehow forgot about the annotation processor generating the manifest. Thanks for the example.

zml2008 (approved):

PR #970, opened 2021-06-12
update gradle version of default wrapper version, 5.6.1 => 7.0.2

keyno63:
I checked to be able to create new project with java16. (by running `gradlew runIde` command)


RedNesto (changes requested):
This breaks fresh Spigot, Bungeecord/Waterfall and SpongeAPI 7 plugins builds.

I plan to update the templates very soon. I recommend you either update the Gradle wrapper for more specific use-cases (like Fabric 1.17 which will eventually be done) or wait for the templates update.

keyno63:
OK. thanks for check this PR.

RedNesto:
The templates update happened a while ago. This PR is no longer relevant.

PR #978, opened 2021-06-15
ats for mojmap

MiniDigger:
this still doesn't detect AT files for paper (even after marking them manually), so goto at doesn't actually work, but copyat does


MiniDigger:
(yes I know this is fucked, no I don't care right now, I was just trying to get copy-at working to save me some time, this is like a shelve, maybe I get some time to redo this cleanly eventually)

PR #994, opened 2021-06-29
General templates cleanup

### LINKED ISSUE #929: Cleanup and update project creator templates
I noticed when working on #928 that the current templates are getting a bit dusty:
- we are using Gradle 5 by default, Gradle 7 has been released for almost 2 months by the time of writing this
- many Gradle buildscripts use deprecated or non-optimal stuff
  - the `maven` plugin was removed
  - `processResources` tasks should use `expand` instead of `from(...) {filter... }` and specify `inputs.properties`
  - use the `plugins` block as much as possible (noticed this with the shadow plugin in multi-project setups)
  - upgrade plugins (we still use shadow 4 while on Gradle 5); SpongeGradle might have received updates since then too
  - other things ?
- Forge 1.12.2 projects still use ForgeGradle 2 while FG 4 can be used (thanks to the amazing work of the retrogradle project contributors)
  - it might be even better to drop 1.12, that's something to decide
- the velocity macros can be used a bit more efficiently, for example we can replace `#if (${HAS_AUTHORS})` by just checking `#if (${AUTHORS})` and set its value to `null` in the context.

I noticed all of this while working on the SpongeAPI 8 plugin template and plan to work on this after getting it merged, but if someone wants to tackle this task let me know.

### LINKED ISSUE #936: New project fails to create with Java 16 and Gradle
* Minecraft Development for IntelliJ plugin version: 2021.2-1.5.11
* IntelliJ version: 2021.2 EAP (212.3116.43)
* Operating System: Windows w/ WSL; IntelliJ is on Windows, Java & Gradle are on Linux

I am running Java 16.0.1 (AdoptOpenJDK-16.0.1+9) and I attempted to create a new project with this plugin on IntelliJ 2021.2 EAP. Everything mostly works, but this comes up:
```
FAILURE: Build failed with an exception.

* What went wrong:
Could not open cp_init remapped class cache for 8lyebhvpycvp55bdjq68eafm6 (/home/simple/.gradle/caches/5.6.1/scripts-remapped/ijtgtmapper1_7qje24re1bfm1hrl1a0070c19/8lyebhvpycvp55bdjq68eafm6/cp_inite7ca86eb00f9d6cbee2b842ef043afcb).
> Could not open cp_init generic class cache for initialization script '\\wsl$\Arch\home\simple\workspace\minecraft\MySecondPlugin\.gradle\ijtgtmapper1.gradle' (/home/simple/.gradle/caches/5.6.1/scripts/8lyebhvpycvp55bdjq68eafm6/cp_init/cp_inite7ca86eb00f9d6cbee2b842ef043afcb).
   > Could not initialize class org.codehaus.groovy.classgen.Verifier

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org
```

I can fix it by simply updating gradle:
`gradle wrapper --gradle-version 7.0.2 --distribution-type all`

I believe then that this is related to #929. Let me know if I can provide more information.

### LINKED ISSUE #973: processResources error
Hello!
With gradle 7.0, if you write it this way, it will throw an error in processResources and will not build.

build.gradle
```
processResources {
    from(sourceSets.main.resources.srcDirs) {
        filter ReplaceTokens, tokens: [version: version]
    }
}
```

error
```
Execution failed for task ':processResources'.
> Entry plugin.yml is a duplicate but no duplicate handling strategy has been set. Please refer to https://docs.gradle.org/7.0.2/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy for details.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
```

### LINKED ISSUE #979: Generate Fabric Mod don't work
The generated fabric mods do not work because of duplicate files caused by:

```

processResources {
    inputs.property "version", project.version

    from(sourceSets.main.resources.srcDirs) {
        include "fabric.mod.json"
        expand "version": project.version
    }

    from(sourceSets.main.resources.srcDirs) {
        exclude "fabric.mod.json"
    }
}
```

RedNesto:
See individual commits for details.

Fixes #929 fixes #936 fixes #973 fixes #979

### DIFF
--- a/src/main/kotlin/creator/buildsystem/gradle/GradleBuildSystem.kt
+++ b/src/main/kotlin/creator/buildsystem/gradle/GradleBuildSystem.kt
@@ -86,7 +86,7 @@ class GradleBuildSystem(
     }
 
     companion object {
-        val DEFAULT_WRAPPER_VERSION = SemanticVersion.release(5, 6, 1)
+        val DEFAULT_WRAPPER_VERSION = SemanticVersion.release(7, 1)
     }
 }
 
--- a/src/main/kotlin/creator/buildsystem/gradle/GradleBuildSystem.kt
+++ b/src/main/kotlin/creator/buildsystem/gradle/GradleBuildSystem.kt
@@ -103,3 +103,9 @@ interface GradleCreator {
      */
     fun configureRootGradle(rootDirectory: Path, buildSystem: GradleBuildSystem) {}
 }
+
+data class GradlePlugin(
+    val id: String,
+    val version: String? = null,
+    val apply: Boolean = true,
+)
--- a/src/main/kotlin/creator/buildsystem/gradle/GradleBuildSystem.kt
+++ b/src/main/kotlin/creator/buildsystem/gradle/GradleBuildSystem.kt
@@ -86,7 +86,7 @@ class GradleBuildSystem(
     }
 
     companion object {
-        val DEFAULT_WRAPPER_VERSION = SemanticVersion.release(7, 1)
+        val DEFAULT_WRAPPER_VERSION = SemanticVersion.release(7, 1, 1)
     }
 }
 
--- a/src/main/kotlin/platform/fabric/creator/FabricProjectSettingsWizard.kt
+++ b/src/main/kotlin/platform/fabric/creator/FabricProjectSettingsWizard.kt
@@ -21,6 +21,7 @@ import com.demonwav.mcdev.platform.fabric.EntryPoint
 import com.demonwav.mcdev.platform.fabric.util.FabricConstants
 import com.demonwav.mcdev.platform.forge.inspections.sideonly.Side
 import com.demonwav.mcdev.util.License
+import com.demonwav.mcdev.util.MinecraftVersions
 import com.demonwav.mcdev.util.SemanticVersion
 import com.demonwav.mcdev.util.modUpdateStep
 import com.demonwav.mcdev.util.toPackageName
@@ -236,6 +237,7 @@ class FabricProjectSettingsWizard(private val creator: MinecraftProjectCreator) 
         conf.mcVersion = mcVersion ?: ""
         val normalizedMcVersion = dataProvider?.getNormalizedMinecraftVersion(mcVersion)?.normalized
         conf.semanticMcVersion = normalizedMcVersion?.let { SemanticVersion.parse(it) } ?: SemanticVersion.release()
+        conf.javaVersion = if (conf.semanticMcVersion >= MinecraftVersions.MC1_17) 16 else 8
         val loaderVer = loaderVersion
         if (loaderVer != null) {
             conf.loaderVersion = SemanticVersion.parse(loaderVer)
@@ -250,11 +252,15 @@ class FabricProjectSettingsWizard(private val creator: MinecraftProjectCreator) 
         conf.gradleVersion = when (dataProvider?.loomVersions?.firstOrNull { it.name == loomVersion }?.gradle) {
             4 -> SemanticVersion.release(4, 10, 3)
             5 -> SemanticVersion.release(5, 6, 4)
-            else -> SemanticVersion.release(6, 5, 1)
+            else -> SemanticVersion.release(6, 9)
         }
         val loomVer = loomVersion
         if (loomVer != null) {
             conf.loomVersion = SemanticVersion.parse(loomVer)
+        }
+        if (conf.loomVersion >= SemanticVersion.release(0, 8)) {
+            // TemplateMakerFabric incorrectly indicates loom 0.8 requires Gradle 6...
+            conf.gradleVersion = SemanticVersion.release(7, 1)
         }
         conf.environment = when ((environmentBox.selectedItem as? String)?.toLowerCase(Locale.ROOT)) {
             "client" -> Side.CLIENT
--- a/src/main/kotlin/platform/fabric/creator/FabricProjectSettingsWizard.kt
+++ b/src/main/kotlin/platform/fabric/creator/FabricProjectSettingsWizard.kt
@@ -258,9 +258,9 @@ class FabricProjectSettingsWizard(private val creator: MinecraftProjectCreator) 
         if (loomVer != null) {
             conf.loomVersion = SemanticVersion.parse(loomVer)
         }
-        if (conf.loomVersion >= SemanticVersion.release(0, 8)) {
+        if (conf.loomVersion >= SemanticVersion.release(0, 7)) {
             // TemplateMakerFabric incorrectly indicates loom 0.8 requires Gradle 6...
-            conf.gradleVersion = SemanticVersion.release(7, 1)
+            conf.gradleVersion = SemanticVersion.release(7, 1, 1)
         }
         conf.environment = when ((environmentBox.selectedItem as? String)?.toLowerCase(Locale.ROOT)) {
             "client" -> Side.CLIENT
--- a/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit Submodule build.gradle.ft
+++ b/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit Submodule build.gradle.ft
@@ -6,7 +6,7 @@ repositories {
 }
 
 dependencies {
-    compile project(':${COMMON_PROJECT_NAME}')
+    api project(':${COMMON_PROJECT_NAME}')
 }
 
 processResources {
--- a/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit Submodule build.gradle.ft
+++ b/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit Submodule build.gradle.ft
@@ -13,6 +13,10 @@ processResources {
     def props = [version: project.version]
     inputs.properties props
     expand props
+    filteringCharset 'UTF-8'
+    filesMatching('plugin.yml') {
+        expand props
+    }
 }
 
 shadowJar {
--- a/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit Submodule build.gradle.ft
+++ b/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit Submodule build.gradle.ft
@@ -10,9 +10,9 @@ dependencies {
 }
 
 processResources {
-    from(sourceSets.main.resources.srcDirs) {
-        expand 'version': project.version
-    }
+    def props = [version: project.version]
+    inputs.properties props
+    expand props
 }
 
 shadowJar {
--- a/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
+++ b/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
@@ -1,5 +1,3 @@
-import org.apache.tools.ant.filters.ReplaceTokens
-
 plugins {
     id 'java'
 }
@@ -18,7 +16,7 @@ dependencies {
 }
 
 processResources {
-    from(sourceSets.main.resources.srcDirs) {
-        filter ReplaceTokens, tokens: [version: version]
-    }
+    def props = [version: version]
+    inputs.properties props
+    expand props
 }
--- a/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
+++ b/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
@@ -19,5 +19,8 @@ dependencies {
 processResources {
     def props = [version: version]
     inputs.properties props
-    expand props
+    filteringCharset 'UTF-8'
+    filesMatching('plugin.yml') {
+        expand props
+    }
 }
--- a/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
+++ b/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
@@ -5,15 +5,27 @@ plugins {
 group = '${GROUP_ID}'
 version = '${PLUGIN_VERSION}'
 
-java {
-    toolchain.languageVersion = JavaLanguageVersion.of(8)
-}
-
 repositories {
     mavenCentral()
 }
 
 dependencies {
+}
+
+def targetJavaVersion = 8
+java {
+    def javaVersion = JavaVersion.toVersion(targetJavaVersion)
+    sourceCompatibility = javaVersion
+    targetCompatibility = javaVersion
+    if (JavaVersion.current() < javaVersion) {
+        toolchain.languageVersion = JavaLanguageVersion.of(targetJavaVersion)
+    }
+}
+
+tasks.withType(JavaCompile).configureEach {
+    if (targetJavaVersion >= 10 || JavaVersion.current().isJava10Compatible()) {
+        options.release = targetJavaVersion
+    }
 }
 
 processResources {
--- a/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
+++ b/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
@@ -5,8 +5,10 @@ plugins {
 group = '${GROUP_ID}'
 version = '${PLUGIN_VERSION}'
 
-sourceCompatibility = '1.8'
-targetCompatibility = '1.8'
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
 
 repositories {
     mavenCentral()
--- a/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
+++ b/src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
@@ -6,8 +6,7 @@ group = '${GROUP_ID}'
 version = '${PLUGIN_VERSION}'
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+    toolchain.languageVersion = JavaLanguageVersion.of(8)
 }
 
 repositories {
--- a/src/main/resources/fileTemplates/j2ee/bungeecord/BungeeCord Submodule build.gradle.ft
+++ b/src/main/resources/fileTemplates/j2ee/bungeecord/BungeeCord Submodule build.gradle.ft
@@ -1,10 +1,12 @@
-apply plugin: 'com.github.johnrengelman.shadow'
+plugins {
+    id 'com.github.johnrengelman.shadow'
+}
 
 repositories {
 }
 
 dependencies {
-    compile project(':${COMMON_PROJECT_NAME}')
+    api project(':${COMMON_PROJECT_NAME}')
 }
 
 processResources {
--- a/src/main/resources/fileTemplates/j2ee/bungeecord/BungeeCord Submodule build.gradle.ft
+++ b/src/main/resources/fileTemplates/j2ee/bungeecord/BungeeCord Submodule build.gradle.ft
@@ -13,6 +13,10 @@ processResources {
     def props = [version: project.version]
     inputs.properties props
     expand props
+    filteringCharset 'UTF-8'
+    filesMatching('bungee.yml') {
+        expand props
+    }
 }
 
 shadowJar {
--- a/src/main/resources/fileTemplates/j2ee/bungeecord/BungeeCord Submodule build.gradle.ft
+++ b/src/main/resources/fileTemplates/j2ee/bungeecord/BungeeCord Submodule build.gradle.ft
@@ -8,9 +8,9 @@ dependencies {
 }
 
 processResources {
-    from(sourceSets.main.resources.srcDirs) {
-        expand 'version': project.version
-    }
+    def props = [version: project.version]
+    inputs.properties props
+    expand props
 }
 
 shadowJar {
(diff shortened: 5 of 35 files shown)


ghost (approved):

src/main/kotlin/platform/fabric/creator/FabricProjectSettingsWizard.kt
```diff
@@ -250,12 +252,16 @@ class FabricProjectSettingsWizard(private val creator: MinecraftProjectCreator)
...
-            else -> SemanticVersion.release(6, 5, 1)
+            else -> SemanticVersion.release(6, 9)
         }
         val loomVer = loomVersion
         if (loomVer != null) {
             conf.loomVersion = SemanticVersion.parse(loomVer)
         }
+        if (conf.loomVersion >= SemanticVersion.release(0, 8)) {
+            // TemplateMakerFabric incorrectly indicates loom 0.8 requires Gradle 6...
+            conf.gradleVersion = SemanticVersion.release(7, 1)
```
zml2008: It seems like `TemplateMakerFabric` has some very shaky logic here: https://github.com/ExtraCrafTX/TemplateMakerFabric/blob/47f1faf0a615612faeb7a1ede6ef08b392d0153b/src/main/java/com/extracraftx/minecraft/templatemakerfabric/data/holders/LoomVersion.java#L13

Really, loom 0.7+ should all default to Gradle 7 for the maximum compatibility

src/main/resources/fileTemplates/j2ee/bukkit/Bukkit Submodule build.gradle.ft
```diff
@@ -6,13 +6,13 @@ repositories {
...
+    api project(':${COMMON_PROJECT_NAME}')
 }
 
 processResources {
-    from(sourceSets.main.resources.srcDirs) {
-        expand 'version': project.version
-    }
+    def props = [version: project.version]
+    inputs.properties props
+    expand props
```
zml2008, line 15: Does it still make sense to apply this to *every* file in the resources directory, or just `plugin.yml` -- applying the transformation to other files could corrupt any binary files shipped with a plugin

You should also explicitly specify the filtering encoding here as:
```
filteringCharset = "UTF-8"
```

for users on Windows

src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
```diff
@@ -1,14 +1,13 @@
...
     id 'java'
 }
 
 group = '${GROUP_ID}'
 version = '${PLUGIN_VERSION}'
 
-sourceCompatibility = '1.8'
-targetCompatibility = '1.8'
+java {
+    toolchain.languageVersion = JavaLanguageVersion.of(8)
```
zml2008: This will, by default, forcibly download a JDK 8 distribution rather than using any newer version available on the user's computer, which is not great for bandwidth and disk space use. The general pattern i'd recommend is something like:

```gradle
def targetVersion = 8 // change this
java {
    def asVersion = JavaVersion.toVersion(targetVersion)
    sourceCompatibility = asVersion
    targetCompatibility = asVersion
    if (JavaVersion.current() < asVersion) {
        toolchain.languageVersion = JavaLanguageVersion.of(targetVersion)
    }
}

tasks.withType(JavaCompile).configureEach {
    if (targetVersion >= 10 || JavaVersion.current().isJava10Compatible()) {
        options.release = targetVersion
    }
}
```

This can be adjusted to any target version, and will ensure that the project will build correctly on any machine with JDK 8 or newer (setting aside compiler bugs in older versions of `javac`)

Ideally `sourceCompatibility` and `targetCompatibility` could be omitted, but it sounds like VS Code depends on those variables to determine the project's java version.

I imagine that with 1.17 using Java 16, the target version should be made a template parameter in the project builder as well, since the majority of new projects will want to be built on Java 16.

src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
```diff
@@ -18,7 +17,7 @@ dependencies {
 }
 
 processResources {
-    from(sourceSets.main.resources.srcDirs) {
-        filter ReplaceTokens, tokens: [version: version]
-    }
+    def props = [version: version]
+    inputs.properties props
+    expand props
```
zml2008: see earlier comment on resources

src/main/resources/fileTemplates/j2ee/bungeecord/BungeeCord Submodule build.gradle.ft
```diff
@@ -1,16 +1,18 @@
...
+    api project(':${COMMON_PROJECT_NAME}')
 }
 
 processResources {
-    from(sourceSets.main.resources.srcDirs) {
-        expand 'version': project.version
-    }
+    def props = [version: project.version]
+    inputs.properties props
+    expand props
```
zml2008, line 15: see earlier comment on resources

src/main/resources/fileTemplates/j2ee/bungeecord/BungeeCord build.gradle.ft
```diff
@@ -1,12 +1,11 @@
...
 apply plugin: 'java'
 
 group = '${GROUP_ID}'
 version = '${PLUGIN_VERSION}'
 
-sourceCompatibility = '1.8'
-targetCompatibility = '1.8'
+java {
+    toolchain.languageVersion = JavaLanguageVersion.of(8)
+}
```
zml2008: see earlier comment on versioning

src/main/resources/fileTemplates/j2ee/bungeecord/BungeeCord build.gradle.ft
```diff
@@ -16,7 +15,7 @@ dependencies {
 }
 
 processResources {
-    from(sourceSets.main.resources.srcDirs) {
-        filter ReplaceTokens, tokens: [version: version]
-    }
+    def props = [version: version]
+    inputs.properties props
+    expand props
```
zml2008, line 32: see earlier comment on resource processing

src/main/resources/fileTemplates/j2ee/fabric/fabric_build.gradle.ft
```diff
@@ -28,33 +32,37 @@ dependencies {
 processResources {
     inputs.property "version", project.version
 
-    from(sourceSets.main.resources.srcDirs) {
-        include "fabric.mod.json"
+    filesMatching("fabric.mod.json") {
         expand "version": project.version
```
zml2008, line 37: a filtering charset should be applied here too

src/main/resources/fileTemplates/j2ee/fabric/fabric_build.gradle.ft
```diff
@@ -28,33 +32,37 @@ dependencies {
...
+tasks.withType(JavaCompile).configureEach {
+    // ensure that the encoding is set to UTF-8, no matter what the system default is
+    // this fixes some edge cases with special characters not displaying correctly
+    // see http://yodaconditions.net/blog/fix-for-java-file-encoding-problems-with-gradle.html
+    // If Javadoc is generated, this must be specified in that task too.
+    it.options.encoding = "UTF-8"
+    #if (${JAVA_VERSION} >= 16)
+
+    // Minecraft 1.17 (21w19a) upwards uses Java 16.
+    it.options.release = 16
```
zml2008: This should be applied on non-16 Java versions as well -- see the snippet above for something that works

src/main/resources/fileTemplates/j2ee/fabric/fabric_build.gradle.ft
```diff
@@ -28,33 +32,37 @@ dependencies {
...
 }
 
-// Loom will automatically attach sourcesJar to a RemapSourcesJar task and to the "build" task
-// if it is present.
-// If you remove this task, sources will not be generated.
-task sourcesJar(type: Jar, dependsOn: classes) {
-    classifier = "sources"
-    from sourceSets.main.allSource
+java {
+    toolchain.languageVersion = JavaLanguageVersion.of(${JAVA_VERSION})
```
zml2008: bandwidth/disk space considerations apply here, as discussed in the initial java versions comment.

(note that toolchains were only added in 6.7, so users on older Loom versions might not be able to use this -- though that begs the question: why even offer pre-0.7 Loom versions to users in the template generator?)

src/main/resources/fileTemplates/j2ee/fabric/fabric_mixins.json.ft
```diff
@@ -2,7 +2,11 @@
   "required": true,
   "minVersion": "0.8",
   "package": "${PACKAGE_NAME}",
+  #if (${JAVA_VERSION} < 16)
   "compatibilityLevel": "JAVA_8",
+  #else
+  "compatibilityLevel": "JAVA_16",
+  #end
```
zml2008:
```suggestion
  "compatibilityLevel": "JAVA_${JAVA_VERSION}",
```

maybe?

src/main/resources/fileTemplates/j2ee/fabric/fabric_submodule_build.gradle.ft
```diff
@@ -26,33 +29,37 @@ dependencies {
...
-// see http://yodaconditions.net/blog/fix-for-java-file-encoding-problems-with-gradle.html
-tasks.withType(JavaCompile) {
-    options.encoding = "UTF-8"
+tasks.withType(JavaCompile).configureEach {
+    // ensure that the encoding is set to UTF-8, no matter what the system default is
+    // this fixes some edge cases with special characters not displaying correctly
+    // see http://yodaconditions.net/blog/fix-for-java-file-encoding-problems-with-gradle.html
+    // If Javadoc is generated, this must be specified in that task too.
+    it.options.encoding = "UTF-8"
+    #if (${JAVA_VERSION} >= 16)
```
zml2008: as above

src/main/resources/fileTemplates/j2ee/fabric/fabric_submodule_build.gradle.ft
```diff
@@ -26,33 +29,37 @@ dependencies {
...
 }
 
-// Loom will automatically attach sourcesJar to a RemapSourcesJar task and to the "build" task
-// if it is present.
-// If you remove this task, sources will not be generated.
-task sourcesJar(type: Jar, dependsOn: classes) {
-    classifier = "sources"
-    from sourceSets.main.allSource
+java {
+    toolchain.languageVersion = JavaLanguageVersion.of(${JAVA_VERSION})
```
zml2008: as above

src/main/resources/fileTemplates/j2ee/sponge/Sponge 8+ Submodule build.gradle.kts.ft
```diff
@@ -55,8 +55,7 @@ sponge {
 
 val javaTarget = 8 // Sponge targets a minimum of Java 8
 java {
-    sourceCompatibility = JavaVersion.toVersion(javaTarget)
-    targetCompatibility = JavaVersion.toVersion(javaTarget)
+    toolchain.languageVersion.set(JavaLanguageVersion.of(javaTarget))
```
zml2008: See the api-9 toolchain logic here: https://github.com/SpongePowered/sponge-plugin-template/blob/api-9/build.gradle.kts#L42, it can be applied to anything 8+ by just changing the Java version number.

src/main/resources/fileTemplates/j2ee/multi/Multi-Module Base build.gradle.ft
```diff
@@ -1,25 +1,20 @@
...
 
-    group = project.group
-    version = project.version
+    group = parent.group
+    version = parent.version
 
-    sourceCompatibility = '1.8'
-    targetCompatibility = '1.8'
+    java {
+        toolchain.languageVersion = JavaLanguageVersion.of(8)
```
zml2008: See java version discussion

src/main/resources/fileTemplates/j2ee/sponge/Sponge 8+ build.gradle.kts.ft
```diff
@@ -53,8 +53,7 @@ sponge {
 
 val javaTarget = 8 // Sponge targets a minimum of Java 8
 java {
-    sourceCompatibility = JavaVersion.toVersion(javaTarget)
-    targetCompatibility = JavaVersion.toVersion(javaTarget)
+    toolchain.languageVersion.set(JavaLanguageVersion.of(javaTarget))
```
zml2008: See above

src/main/resources/fileTemplates/j2ee/velocity/Velocity Submodule build.gradle.ft
```diff
@@ -18,12 +17,11 @@ shadowJar {
...
 
-import org.apache.tools.ant.filters.ReplaceTokens
-
 task processSources(type: Sync) {
     from sourceSets.main.java.srcDirs
-    inputs.property 'version', version
-    filter ReplaceTokens, tokens: [version: version]
+    def props = [version: project.version]
+    inputs.properties props
+    expand props
```
zml2008: Comment about resource filtering applies here too.

The way this is set up for source filtering is a bit strange too -- it seems like it would lead to compile errors not linking to the correct location in the IDE.

What I usually do is generate a separate package-private source file with constants, and register that as a synchronization task.

See an example [in VanillaGradle](https://github.com/SpongePowered/VanillaGradle/blob/7548833f611dc9b856197e4a3855865918c69d37/build.gradle.kts#L102), which processes https://github.com/SpongePowered/VanillaGradle/blob/7548833f611dc9b856197e4a3855865918c69d37/src/main/templates/org/spongepowered/gradle/vanilla/internal/BuildVersions.java#L35

src/main/resources/fileTemplates/j2ee/velocity/Velocity build.gradle.ft
```diff
@@ -5,8 +5,9 @@ plugins {
 group = '${GROUP_ID}'
 version = '${PLUGIN_VERSION}'
 
-sourceCompatibility = '1.8'
-targetCompatibility = '1.8'
+java {
+    toolchain.languageVersion = JavaLanguageVersion.of(8)
+}
```
zml2008: see comment about java versions

src/main/resources/fileTemplates/j2ee/velocity/Velocity build.gradle.ft
```diff
@@ -15,12 +16,11 @@ repositories {
...
-
 task processSources(type: Sync) {
     from sourceSets.main.java.srcDirs
-    inputs.property 'version', version
-    filter ReplaceTokens, tokens: [version: version]
+    def props = [version: version]
+    inputs.properties props
+    expand props
     into "$buildDir/src"
 }
```
zml2008: see above

zml2008 (changes requested):
A few things here -- I've proposed some alternatives to choices you make across your templates, but you'll have to balance how correct the buildscripts are with how much you want to keep the buildscripts simple for users new to Gradle

zml2008:
Another little thing that might be good to include: [this block](https://github.com/SpongePowered/sponge-plugin-template/blob/5d1ad3ed7f9b245aa05596b153c438c37592f397/build.gradle.kts#L55) that makes builds more reproducible

src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
DenWav: Would rather not use `sourceCompatibility` and `targetCompatibility` for Java 9+. I'm not concerned about VSCode compatibility for two reasons:

 1. This is an IntelliJ plugin
 2. We should always focus on a proper Gradle setup, rather than doing not great things for the sake of tools which don't support Gradle correctly.

In short, that sounds like a VSCode problem, not our problem.

DenWav:
> Another little thing that might be good to include: this block that makes builds more reproducible

Oh I wish that was default. Didn't know about these options, will probably start adding that to all of my builds. Thanks! Also lots of info overall, thanks for the review.

src/main/resources/fileTemplates/j2ee/fabric/fabric_mixins.json.ft
DenWav: Didn't know that you could just have the number in the Mixin config. That is a lot cleaner.

src/main/kotlin/creator/buildsystem/gradle/GradleBuildSystem.kt
```diff
@@ -86,7 +86,7 @@ class GradleBuildSystem(
     }
 
     companion object {
-        val DEFAULT_WRAPPER_VERSION = SemanticVersion.release(5, 6, 1)
+        val DEFAULT_WRAPPER_VERSION = SemanticVersion.release(7, 1)
```
DenWav: Should bump this to 7.1.1 now.

We should make this into a data file like the platform versions though, so it doesn't require a code change. Out of scope for thiw PR unless you want to do it.

src/main/resources/fileTemplates/j2ee/sponge/Sponge build.gradle.ft
```diff
@@ -5,8 +5,9 @@ plugins {
 group = '${GROUP_ID}'
 version = '${PLUGIN_VERSION}'
 
-sourceCompatibility = '1.8'
-targetCompatibility = '1.8'
+java {
+    toolchain.languageVersion.set(JavaLanguageVersion.of(8))
```
DenWav: This applies to each of the changes in the toolchains commit - Just making sure that there is no way a user could configure a project setup which would lead to one of these templates to be applied to a version of Gradle < 6.7?

DenWav (changes requested):
Lots of good changes. Thanks @zml2008 for the review, really helpful and a lot of good information.

src/main/resources/fileTemplates/j2ee/fabric/fabric_mixins.json.ft
zml2008: oops, it's not just the java version, i just gave an incorrect correction. I've updated the original comment.

The compatibility level system is a bit more complex than just matching the bytecode version, but I think matching the bytecode version is a fair start for now (internally as of 0.8.3, mixin traverses each class file to determine which bytecode features are used to validate the compatibility level, and I expect more improvements over time)

src/main/resources/fileTemplates/j2ee/bukkit/Bukkit build.gradle.ft
zml2008: That's fair, though for 1 -- while this is an intellij plugin, whoever creates the project using the mc dev templates is not necessarily the only contributor, so buildscripts should be reasonably consumable by any common IDE. I think you could reasonably say that VS Code is not a common Java IDE though :p

RedNesto: > This will, by default, forcibly download a JDK 8 distribution rather than using any newer version available on the user's computer

Gradle will first search for existing installations (I expect most plugin/mod developer to have installed a JDK 8 by standard means.) For the ones that care about that stuff they can even tell Gradle about their custom installations: https://docs.gradle.org/current/userguide/toolchains.html#sec:custom_loc

It all depends if you want the most reproducible builds possible (in that case we would also set a vendor and implementation.)

I guess your solution is the best for the average developer experience.

src/main/kotlin/platform/fabric/creator/FabricProjectSettingsWizard.kt
RedNesto: I will lower the condition to loom 0.7 then.

src/main/resources/fileTemplates/j2ee/fabric/fabric_build.gradle.ft
RedNesto: I know close to nothing about the fabric ecosystem, if we can filter out older loom versions without preventing people from developing on supported versions then that is something I can consider.

src/main/resources/fileTemplates/j2ee/velocity/Velocity Submodule build.gradle.ft
RedNesto: I have to admit this is a tricky one because we use velocity's AP to generate the plugin manifest, and it gets the version from the plugin annotation (like with SpongeAPI 7 plugins.)

Maybe we can expand the generated manifest instead of the source file ?

For reference [this is where we generate the version attribute of the plugin annotation](https://github.com/minecraft-dev/MinecraftDev/blob/d8d0e4eaa682ee07db1471cbe5d3e168a811db93/src/main/kotlin/platform/velocity/creator/VelocityProjectCreator.kt#L177).

src/main/kotlin/creator/buildsystem/gradle/GradleBuildSystem.kt
RedNesto: We might have to rethink Gradle versions handling in the wizard as a whole I think. The old SpongeGradle plugin for SpongeAPI 7 doesn't work on newer Gradle versions for example.

Same for Java versions, we could set a min/max Java version in each project config, and warn the user about conflicts (max bound in one config lower than lower bound of another config.) Clearly out of the scope if this PR though ^^

src/main/resources/fileTemplates/j2ee/sponge/Sponge build.gradle.ft
RedNesto: I have to test that, it might be an issue with FG2 projects.

PR #995, opened 2021-07-02
ForgeGradle 5 support

RedNesto:
A ForgeGradle 5 broke the Gradle tooling extension some time ago, not too big though: 2 return types changed.

I chose the simple path and added simple instanceof checks to the FG3 model builder and assume there will be no more breaking changes for FG5.

This PR might sit until FG5 is stabilized, we'll see how it goes.


SizableShrimp:
I doubt we will be changing these return types any further as they were moved to properties and will stay that way. I would like to see this merged so that support for FG5 is back in the plugin.

DenWav (approved):

SizableShrimp:
ForgeGradle 5 has hit RB of 5.1. Which means no more binary breaks on 5.1 and if we need to then we move to 5.2. I would love to see this merged now that FG5 is the default in the mdks of both 1.16 and 1.17

PR #996, opened 2021-07-02
Advertise VanillaGradle way to get MC sources

RedNesto:
VanillaGradle requires the user to run the `decompile` task to get Minecraft sources.

This PR adds an option that runs this task and refreshes the Gradle project when navigating MC decompiled sources. IntelliJ automagically opens the opens the source file in the editor once the refresh completes.

Here's what it looks like currently: 
image

The code is working correctly on single-project and basic multi-project setups (it worked fine on Sponge branch api-8 for example.)

However, this PR does not provide "MCP" data MCDev uses in some places (MC version, mappings, etc...), this is out of the scope of this PR. This is more like a proof-of-concept for this particular feature, rather than complete VanillaGradle support.


DenWav (approved):
Super cool, thanks

PR #997, opened 2021-07-02
Add license field to Forge project creator

RedNesto:
Allows to set the license of a Forge mod from the project creator.

image

The combobox is wider than necessary because some license names are quite long.

The default selected license is `All Rights Reserved`, like in the MDK.

### DIFF
--- a/src/main/kotlin/creator/LicenseStep.kt
+++ b/src/main/kotlin/creator/LicenseStep.kt
@@ -1,0 +1,41 @@
+/*
+ * Minecraft Dev for IntelliJ
+ *
+ * https://minecraftdev.org
+ *
+ * Copyright (c) 2021 minecraft-dev
+ *
+ * MIT License
+ */
+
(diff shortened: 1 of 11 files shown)


Mysterious-Dev (approved):

src/main/kotlin/creator/LicenseStep.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val project: Project,
+    private val rootDirectory: Path,
+    private val license: License,
+    private val author: String
+) : CreatorStep {
+
+    override fun runStep(indicator: ProgressIndicator) {
+        val licenseFile = rootDirectory.resolve("LICENSE")
+
+        val fileText = FabricTemplate.applyLicenseTemplate(project, license, author)
```
DenWav: Maybe we should move this from `FabricTemplate` to a `CommonTemplate` object or something like that.

DenWav (approved):
Looks good.

PR #1506, opened 2021-09-07
Member reference rewrite

Earthcomputer:
- Abstract member reference into mixin selectors
- Add an extension point to allow for custom mixin selectors to be parsed
- Add support for builtin regex and `@Desc` selectors.
- Support for `@Desc` more generally is not included in this PR, and will be added later.


RedNesto (approved):
It looks great. I don't know enough about the selectors to test it properly though.

PR #1525, opened 2021-09-10
Mixin annotation handlers and injection points to extension points, add support for more injection points

Earthcomputer:
Watch your back, mcdev is getting smarter!

- Added an extension point for "annotation handlers", which handle a mixin annotation on a method or a field.
   - Unified some functionality between `@Overwrite`, `@Shadow`, `@Accessor` and `@Inject` using this new system.
   - Added annotation handlers for `@Inject`, `@ModifyArg`, `@ModifyArgs`, `@ModifyVariable` and `@Redirect`, meaning the unresolved target inspection and the mixin target line marker provider also apply to them.
   - There is no support for `@ModifyConstant` yet, although this is planned for the future; the system is designed to support annotations like it.
   - The `isInsnAllowed` method is currently unused, but is planned for a future inspection.
- Added an extension point for injection points, keyed by their AtCode (or class name if there is no at code).
   - Updated the `INVOKE`, `INVOKE_ASSIGN`, `INVOKE_STRING`, `FIELD` and `NEW` injection points to use this system.
   - Added support for `HEAD`, `RETURN` and `TAIL`.
   - Added support for recognizing the target of `NEW` using `args = "class=..."` because that's a thing apparently.
   - There is no support for `LOAD`, `STORE` and `CONSTANT` yet, although this is planned for the future.


Earthcomputer:
Also fixed #1527 in this PR

RedNesto:
I noticed we can't navigate to RETURN targets when injecting into `<init>` or `<clinit>`. The navigation gutter icon is there but clicking it does nothing (for example https://github.com/SpongePowered/Sponge/blob/d528567cbc906a457bc8a1eac520df8245f022e9/src/mixins/java/org/spongepowered/common/mixin/core/world/level/ExplosionMixin.java#L97). Otherwise this looks really good 👍

Earthcomputer:
Oh right, that's the implicit return in void methods, I sort of forgot those existed

Earthcomputer:
It now navigates to the closing brace of the method if an implicit return statement is detected.

RedNesto (approved):

PR #1538, opened 2021-09-12
Move InjectorType functionality to injector handlers

Earthcomputer:
This means that the injector parameters inspection and quick fix is more accurate and supports more injectors.

Also added array opcode support to the field injection point, which is what started this whole rabbit whole in the first place.


Earthcomputer:
Note that there are some false positives that are particularly noticeable on field redirects - this is because the `AtResolver` doesn't yet take into account all available information, resulting in too many matches, so mcdev thinks that it can't find a single method signature to satisfy all of them. This will be fixed once more precise matching is implemented, which I'm planning on next.

RedNesto (approved):

PR #1797, opened 2022-01-20
Add support for architectury api

KP2048:


KP2048:
Hooray for architectury

gradle.properties
```diff
@@ -11,10 +11,10 @@
 # suppress inspection "UnusedProperty" for whole file
 kotlin.code.style=official
 
-ideaVersion = 2021.1
-ideaVersionName = 2021.1
+ideaVersion = 2021.3
+ideaVersionName = 2021.3
```
Earthcomputer: Can you leave these settings alone?

build.gradle.kts
```diff
@@ -153,7 +154,7 @@ tasks.publishPlugin {
 }
 
 tasks.runPluginVerifier {
-    ideVersions.addAll("IC-2020.1.3", "IC-2020.1.4")
+    ideVersions.addAll("IC-2021.3")
```
Earthcomputer: Could you leave these settings alone?

gradle/libs.versions.toml
```diff
@@ -15,8 +15,9 @@ jflex-skeleton = "org.jetbrains.idea:jflex:1.7.0-c1fdf11"
 grammarKit = "org.jetbrains.idea:grammar-kit:1.5.1"
 
 # Gradle Tooling
-gradleToolingExtension = "com.jetbrains.intellij.gradle:gradle-tooling-extension:211.6693.111"
+gradleToolingExtension = "com.jetbrains.intellij.gradle:gradle-tooling-extension:213.5744.223"
```
Earthcomputer: Could you leave these settings alone?

src/main/kotlin/nbt/lang/NbttParserDefinition.kt
```diff
@@ -28,7 +28,7 @@ class NbttParserDefinition : ParserDefinition {
     override fun createFile(viewProvider: FileViewProvider) = NbttFile(viewProvider)
 
     override fun spaceExistenceTypeBetweenTokens(left: ASTNode, right: ASTNode) =
-        LanguageUtil.canStickTokensTogetherByLexer(left, right, NbttLexerAdapter())!!
+        LanguageUtil.canStickTokensTogetherByLexer(left, right, NbttLexerAdapter())
```
Earthcomputer: Can you leave out unrelated changes like this?

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    fun getArchitecturyVersions(mcVersion: SemanticVersion): List<SemanticVersion> {
+        return versions.asSequence()
+            .filter {
+                it.startsWith(
+                    when {
+                        mcVersion < MinecraftVersions.MC1_17 -> "1"
+                        mcVersion >= MinecraftVersions.MC1_17 && mcVersion < MinecraftVersions.MC1_18 -> "2"
+                        mcVersion >= MinecraftVersions.MC1_18 -> "3"
+                        else -> "0"
```
Earthcomputer: If there is an API for this, it would be better to use that so we don't have to maintain this list

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
```diff
@@ -0,0 +1,116 @@
...
+
+                        val versionEvent = reader.next()
+                        if (!versionEvent.isCharacters) {
+                            continue
+                        }
+                        val version = versionEvent.asCharacters().data
+                        // val index = version.indexOf('-')
+                        // if (index == -1) {
+                        //     continue
+                        // }
```
Earthcomputer: Is this comment still needed? If not better remove it

src/main/kotlin/translations/TranslationEditorNotificationProvider.kt
```diff
@@ -44,7 +44,7 @@ class TranslationEditorNotificationProvider :
         val missingTranslations = getMissingTranslations(project, file)
         if (missingTranslations.any()) {
             val panel = InfoPanel()
-            panel.setText("Translation file doesn't match default one (${TranslationConstants.DEFAULT_LOCALE} locale).")
+            panel.text = "Translation file doesn't match default one (${TranslationConstants.DEFAULT_LOCALE} locale)."
```
Earthcomputer: Please refrain from making unrelated changes like this

src/main/kotlin/translations/lang/LangParserDefinition.kt
```diff
@@ -27,7 +27,7 @@ class LangParserDefinition : ParserDefinition {
     override fun createFile(viewProvider: FileViewProvider) = LangFile(viewProvider)
 
     override fun spaceExistenceTypeBetweenTokens(left: ASTNode, right: ASTNode) =
-        LanguageUtil.canStickTokensTogetherByLexer(left, right, LangLexerAdapter())!!
+        LanguageUtil.canStickTokensTogetherByLexer(left, right, LangLexerAdapter())
```
Earthcomputer: More unrelated changes

src/main/resources/META-INF/plugin.xml
```diff
@@ -34,7 +34,7 @@
     </change-notes>
 
     <!-- please see https://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description -->
-    <idea-version since-build="211.4961.33" until-build="211.*"/>
+    <idea-version since-build="213.3714.440" until-build="213.*"/>
```
Earthcomputer: This should not be in the dev branch

Earthcomputer (changes requested):
Finally got the chance to review this. Please remove unrelated changes, they make it hard to review.

KP2048:
The idea version changes were so I could test them / use my fork on my ide… I can change them back if you like

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
KP2048: Short of making a nodejs rest api for parsing the mc version off of the forge jar file, I see no good way of doing this. I don’t want to have to download the jar file and load it into the classpath at runtime and use reflection…

KP2048: Although that rest api could be useful for other people too so I might do it :)

Earthcomputer: It might be best to ask architectury to provide such an API themselves

gradle/libs.versions.toml
```diff
@@ -17,6 +17,7 @@ grammarKit = "org.jetbrains.idea:grammar-kit:1.5.1"
 # Gradle Tooling
 gradleToolingExtension = "com.jetbrains.intellij.gradle:gradle-tooling-extension:211.6693.111"
 annotations = "org.jetbrains:annotations:20.1.0"
+groovy = "org.codehaus.groovy:groovy-all:2.5.14"
```
Earthcomputer: I think the groovy change is breaking the build

Earthcomputer (changes requested):

src/main/kotlin/insight/ColorUtil.kt
```diff
@@ -193,7 +193,7 @@ fun UElement.setColor(color: String) {
     sourcePsi.containingFile.runWriteAction {
         val project = sourcePsi.project
         val parent = this.uastParent
-        val newColorRef = generationPlugin?.getElementFactory(project)?.createQualifiedReference(color, this)
+        val newColorRef = generationPlugin?.getElementFactory(project)?.createQualifiedReference(color, sourcePsi)
```
Earthcomputer: Can you fix all the compile errors related to downgrading intellij again? Such as this one here.

Earthcomputer (changes requested):

src/main/kotlin/insight/ColorUtil.kt
KP2048: Yes, however I’m currently on vacation and don’t have access to my pc so it will have to wait till I have access

KP2048: Used https://gitpod.io

KP2048:
However the nbtt parser test fails as well as the Access Transformer parser tests

src/test/kotlin/framework/test-util.kt
```diff
@@ -96,7 +96,7 @@ fun ProjectBuilderTest.testParser(basePath: String, func: ProjectBuilderFunc) {
         file = func(basePath.substringAfterLast('/'), text, true, false).toPsiFile()
     }
 
-    val actual = DebugUtil.psiToString(file!!, false, true)
+    val actual = DebugUtil.psiToString(file!!, true, true)
```
Earthcomputer: Presumably because of this

KP2048: ah

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
KP2048: I’ve decided. I’m making that rest api because it’s cool and useful. And it would clean this bit up here real nice.

Earthcomputer: It really would be better to have an official API, for a plugin that so many people use it would be more reliable. I've asked about it in their discord.

KP2048: So have I

Earthcomputer:
For reference, here's what the architectury project creator looks like currently:
Screenshot from 2022-03-16 00-03-07

Aren't Fabric API and Architectury API both optional? This GUI doesn't reflect that.

KP2048:
Architectury api requires fabric api so disabling fabric api would need to disable architectury api too

KP2048:
I can add a check box but that’s beyond the limit of what I can do on my phone. The soonest I can do that is Thursday

Earthcomputer:
No rush

Earthcomputer:
You mentioned before that you made changes because you couldn't get 2021.1 to work in development. Usually it's just a case of adding `systemProp.mcdev.localdev=true` to your global `gradle.properties` file; if you're still having issues with it, ask in the mcdev discord and we can get it sorted.

KP2048:
It’s not that I have compile issues, the built plug-in wouldn’t install on my 2021.3 ide. I can test it on the 2021.1 community edition gradle downloads but I wanted to use my fork in my ide so…

Earthcomputer:
Are you using the `runIde` gradle task?

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
KP2048: For future reference and so I don’t forget,
Here’s the link to the json:
https://gist.githubusercontent.com/shedaniel/4a37f350a6e49545347cb798dbfa72b3/raw/architectury.json

KP2048:
Runide gradle task works fine but installing to my IntelliJ IDEA Ultimate 2021.3 says outdated plug-in for some reason even though I can install the marketplace one just fine

Earthcomputer:
In development, you test using runIde. The development branch uses the oldest IDEA version that we support, and we have separate branches for each of the newer versions that we merge into periodically. This is done automatically where possible for the nightly builds.

KP2048:
So once the pr gets merged it will propagate?

Earthcomputer:
Yes

KP2048:
unfortunately architectury common module compile only depends on fabric loader ONLY FOR @Environment annotation. this means that the module system detects the architectury common module as a fabric module and an architectury module and i can't seem to fix that without breaking all fabric modules

Earthcomputer:
Hmm, we could change how Fabric modules are detected to avoid this

KP2048:
im currently changing things to make architectury work in multi module setup

KP2048:
ive also found that the settings.gradle in the submodules doesn't work in multi module setup. 
this results in fabric-loom/architectury-loom not being able to be found in repositories
as those repositories are defined in the settings.gradle in the submodule and gradle only loads the settings.gradle in the root project

Witherking25:
this results in fabric-loom/architectury-loom not being able to be found in repositories

Earthcomputer:
Is this ready now?

KP2048:
almost i just need to do some filtering on the allowed mc versions because architectury doesn't support all of them

Witherking25:
ready

KP2048:
Just added it to the readme :)

KP2048:
Ready

KP2048:
Current ui for project creator:

image

KP2048:
Hrm, maybe i should align things better

KP2048:
That's better:

image

RedNesto:
You will have to drop all the commits that were merged from the release branches before this PR is merged.

KP2048:
how would i go about doing that?

KP2048:
git rebase -i?

KP2048:
Figured it out :)

Earthcomputer:
Looks like the proper file icons aren't showing for me
Screenshot from 2022-03-22 19-10-12
I expected an architectury icon for the common source sets, instead it shows a Minecraft grass block icon.

src/main/kotlin/platform/architectury/creator/ArchitecturyProjectConfig.kt
```diff
@@ -0,0 +1,78 @@
...
+    var forgeVersion: SemanticVersion = SemanticVersion.release()
+    var forgeVersionText: String = ""
+    var fabricLoaderVersion = SemanticVersion.release()
+    var fabricApiVersion: SemanticVersion = SemanticVersion.release()
+    var architecturyApiVersion: SemanticVersion = SemanticVersion.release()
+    var loomVersion = SemanticVersion.release()
+    private var gradleVersion = SemanticVersion.release()
+    val architecturyGroup: String
+        get() = when {
+            mcVersion >= MinecraftVersions.MC1_17 -> "dev.architectury"
```
Earthcomputer: Nit: could this use architectury version instead of mc version?

src/main/kotlin/platform/architectury/creator/ArchitecturyProjectSettingsWizard.kt
```diff
@@ -0,0 +1,318 @@
...
+
+    override fun updateDataModel() {
+        val conf = this.config ?: return
+
+        conf.pluginName = this.modNameField.text
+        conf.description = this.descriptionField.text
+        conf.website = this.websiteField.text
+
+        conf.modRepo = this.repositoryField.text
+        conf.modRepo = this.issueField.text
```
Earthcomputer: Think you're setting the wrong field here

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
```diff
@@ -0,0 +1,144 @@
...
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+
+class ArchitecturyVersion private constructor(val versions: Map<SemanticVersion, List<SemanticVersion>>) {
+
+    fun getArchitecturyVersions(mcVersion: SemanticVersion): List<SemanticVersion> {
+        val roundedVersion = Json.parseToJsonElement(
+            URL(
+                "https://gist.githubusercontent.com" +
```
Earthcomputer: All the URLs in this file could be extracted out into constants

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
```diff
@@ -0,0 +1,144 @@
...
+                    "/shedaniel/4a37f350a6e49545347cb798dbfa72b3" +
+                    "/raw/architectury.json"
+            ).readText()
+        ).jsonObject["versions"]!!.jsonObject.map { SemanticVersion.parse(it.key) }
```
Earthcomputer: If the format changes, we should ideally fail gracefully rather than asserting things are not null.

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
```diff
@@ -0,0 +1,144 @@
...
+                            result[findMcVersion(SemanticVersion.parse(version))]!!.add(SemanticVersion.parse(version))
+                        } else {
+                            result[findMcVersion(SemanticVersion.parse(version))] = mutableListOf(
+                                SemanticVersion.parse(version)
+                            )
+                        }
+                    }
+                }
+                url2.openStream().use { stream ->
+                    val inputFactory = XMLInputFactory.newInstance()
```
Earthcomputer: This code is a duplicate of the code above.

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
```diff
@@ -0,0 +1,144 @@
...
+            ).jsonObject
+            return SemanticVersion.parse(
+                meta["versions"]!!.jsonObject.filter {
+                    it.value.jsonObject["api"]!!.jsonObject["filter"]!!.jsonPrimitive.content.toRegex().matches(
+                        architecturyVersion.toString()
+                    )
+                }.first().key
+            )
+        }
+        fun downloadData(): ArchitecturyVersion? {
```
Earthcomputer: This function (and maybe some others in this file) could be coroutines, which would allow sending these get requests in parallel.

Earthcomputer (changes requested):

src/main/kotlin/platform/architectury/creator/ArchitecturyProjectSettingsWizard.kt
KP2048: oop

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
KP2048: there are 2 urls and they both need the same code.
i guess i could use a loop?

KP2048:
> Looks like the proper file icons aren't showing for me Screenshot from 2022-03-22 19-10-12 I expected an architectury icon for the common source sets, instead it shows a Minecraft grass block icon.

it shows a grass icon because it is being detected as both a fabric module and an architectury module. architectury common module references fabric loader for @Environment annotation, and fabric loader is the criteria for being a fabric module so it gets mistakenly detected as a fabric module. I'd have to rework how fabric modules are detected and thats beyond the scope of adding architectury

if you manually disable fabric for the module it correctly shows the icon

image

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
KP2048: So do a chain of `?.` thingies and check if null at the end?

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
KP2048: technically downloadData() is called in 
https://github.com/minecraft-dev/MinecraftDev/blob/22f04db2502b5f14f8da406dfbe3a61af1967436/src/main/kotlin/platform/architectury/creator/ArchitecturyProjectSettingsWizard.kt#L256-L266
so it is already asynchronus

KP2048: currently this is what i have
```kotlin
class ArchitecturyVersion private constructor(
    val versions: Map<SemanticVersion, List<SemanticVersion>>,
    private val mcVersions: MutableList<List<SemanticVersion>>
) {


    fun getArchitecturyVersions(mcVersion: SemanticVersion): List<SemanticVersion> {
        val roundedVersion = mcVersions.find { mcVersion >= it[0] && mcVersion < it[1] }?.first()
        return try {
            versions[roundedVersion]?.asSequence()
                ?.sortedDescending()
                ?.take(50)
                ?.toList() ?: throw IOException("Could not find any architectury versions for $mcVersion")
        } catch (e: IOException) {
            e.printStackTrace()
            emptyList()
        }
    }

    companion object {
        private val updateUrl = URL(
            "https://gist.githubusercontent.com" +
                    "/shedaniel/4a37f350a6e49545347cb798dbfa72b3" +
                    "/raw/architectury.json"
        )

        fun downloadData(): ArchitecturyVersion? {
            try {
                val meta = Json.parseToJsonElement(updateUrl.readText())
                val versions = mutableMapOf<SemanticVersion, MutableList<SemanticVersion>>()
                val mcVersions = meta.jsonObject["versions"]?.jsonObject?.map { SemanticVersion.parse(it.key) }
                    ?.windowed(2, 1)?.toMutableList().also {
                        it?.add(
                            listOf(
                                it.last().last(),
                                SemanticVersion.parse(
                                    when (val part = it.last().last().parts.getOrNull(1)) {
                                        is SemanticVersion.Companion.VersionPart.ReleasePart -> (part.version + 1).toString()
                                        null -> "?"
                                        else -> part.versionString
                                    }
                                )
                            )
                        )
                    } ?: throw IOException("Could not find any minecraft versions")

                meta.jsonObject["versions"]?.jsonObject?.forEach {
                    val mcVersion = SemanticVersion.parse(it.key)
                    URL(
                        it.value.jsonObject["api"]?.jsonObject?.get("pom")?.jsonPrimitive?.content ?: throw IOException(
                            "Could not find pom for $mcVersion"
                        )
                    )
                        .openStream().use { stream ->
                            val inputFactory = XMLInputFactory.newInstance()

                            @Suppress("UNCHECKED_CAST")
                            val reader = inputFactory.createXMLEventReader(stream) as Iterator<XMLEvent>
                            for (event in reader) {
                                if (!event.isStartElement) {
                                    continue
                                }
                                val start = event.asStartElement()
                                val name = start.name.localPart
                                if (name != "version") {
                                    continue
                                }

                                val versionEvent = reader.next()
                                if (!versionEvent.isCharacters) {
                                    continue
                                }
                                val version = versionEvent.asCharacters().data
                                versions.getOrPut(mcVersion) { mutableListOf() }.add(SemanticVersion.parse(version))

                            }
                        }
                }

                return ArchitecturyVersion(versions, mcVersions)
            } catch (e: IOException) {
                e.printStackTrace()
                return null
            }
        }
    }
}
```

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
KP2048: nah i reworked it and made it better see below

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
Earthcomputer: It is asynchronous to the UI thread but that doesn't mean they're running in parallel to each other. Currently, you're sending a get request for each Minecraft version serially, where you have to wait the length of the ping for any response at all. If the requests are in parallel, in theory you only have to wait the length of the ping once, rather than once for each Minecraft version (which will get worse as more versions are added).

Earthcomputer:
The icons may be wrong because of the Fabric library detection, but it still needs to be fixed. We can't expect users to go in manually and untick Fabric.

KP2048:
fair enough

KP2048:
hrm, when architectury api is disabled, there is no way to determine which one is the common and which one is the fabric...

KP2048:
actually, architectury injectables seems to only be in common and not fabric

KP2048:
hrm, but main not test

KP2048:
if i make fabric module have icon secondary, that would fix it.

KP2048:
also technically architectury api is optional but without it you might as well just use a standard multi module setup. its like saying you technically don't need a modloader because you can just use jarmodding. it just isn't the same. it's not architectury without all the code that bridges the gap between fabric and forge. i also cant detect the test source set of the common module as architectury without it

KP2048:
it has exactly the same libraries as the fabric module's main source set

KP2048:
i mean i could make a dummy library and add that to the dependencies but that is kinda cheating

Earthcomputer:
Take a look at `com.demonwav.mcdev.facet.MinecraftFacetDetector.MinecraftModuleRootListener#autoDetectTypes`. You probably want to achieve this by adding custom logic to that method.

KP2048:
image

Earthcomputer:
The problem with that approach is that it puts the architectury icon on any module named "common". I have managed to get something working locally that instead detects it via testing for the architectury gradle plugin and the presence of a gradle task.

KP2048:
Ok pr?

Earthcomputer:
https://github.com/withertech/MinecraftDev/pull/2

Earthcomputer:
Weirdly it puts the icon on the whole module as well as the source sets, but afaik that's the intended behaviour and everything else isn't working, so I'll leave it like this for now, we can fix the other platforms later.

build.gradle.kts
```diff
@@ -298,7 +298,15 @@ license {
 ktlint {
     enableExperimentalRules.set(true)
 }
-
+// fun String.execute(currentWorkingDir: File = file("./")): InputStreamReader {
```
Earthcomputer: This commented out code isn't needed anymore, right?

Earthcomputer (changes requested):
We're getting close to this being mergeable now. Just need to resolve [this](https://github.com/minecraft-dev/MinecraftDev/pull/1797#discussion_r833729788), and one or two more things:

build.gradle.kts
KP2048: Nah

src/main/kotlin/platform/architectury/version/ArchitecturyVersion.kt
KP2048:
```kotlin
      meta.jsonObject["versions"]?.jsonObject?.asSequence()?.asStream()?.parallel()?.forEach {

```

build.gradle.kts
KP2048: nope

Earthcomputer:
Can you do it using coroutines? That way they're on the same executor as the other download tasks.

KP2048:
```kotlin
suspend fun downloadData(): ArchitecturyVersion? = coroutineScope {
            try {
                val meta = withContext(Dispatchers.IO) { Json.parseToJsonElement(updateUrl.readText()) }
                val versions = mutableMapOf<SemanticVersion, MutableList<SemanticVersion>>()
                val mcVersions = meta.jsonObject["versions"]?.jsonObject?.map { SemanticVersion.parse(it.key) }
                    ?.windowed(2, 1)?.toMutableList().also {
                        it?.add(
                            listOf(
                                it.last().last(),
                                SemanticVersion.parse(
                                    buildString {
                                        append("1.")
                                        append(
                                            when (val part = it.last().last().parts.getOrNull(1)) {
                                                is SemanticVersion.Companion.VersionPart.ReleasePart ->
                                                    (part.version + 1).toString()
                                                null -> "?"
                                                else -> part.versionString
                                            }
                                        )
                                    }
                                )
                            )
                        )
                    } ?: throw IOException("Could not find any minecraft versions")
                meta.jsonObject["versions"]?.jsonObject?.asSequence()?.map  {
                    async(Dispatchers.IO) {
                        val mcVersion = SemanticVersion.parse(it.key)
                        URL(
                            it.value.jsonObject["api"]?.jsonObject?.get("pom")?.jsonPrimitive?.content
                                ?: throw IOException(
                                    "Could not find pom for $mcVersion"
                                )
                        )
                            .openStream().use { stream ->
                                val inputFactory = XMLInputFactory.newInstance()

                                @Suppress("UNCHECKED_CAST")
                                val reader = inputFactory.createXMLEventReader(stream) as Iterator<XMLEvent>
                                for (event in reader) {
                                    if (!event.isStartElement) {
                                        continue
                                    }
                                    val start = event.asStartElement()
                                    val name = start.name.localPart
                                    if (name != "version") {
                                        continue
                                    }

                                    val versionEvent = reader.next()
                                    if (!versionEvent.isCharacters) {
                                        continue
                                    }
                                    val version = versionEvent.asCharacters().data
                                    val regex = it.value.jsonObject["api"]?.jsonObject?.get("filter")
                                        ?.jsonPrimitive?.content?.toRegex()
                                        ?: throw IOException("Could not find filter for $mcVersion")
                                    if (regex.matches(version)) {
                                        versions.getOrPut(mcVersion) { mutableListOf() }
                                            .add(SemanticVersion.parse(version))
                                    }
                                }
                            }
                    }
                }?.asSequence()?.toList()?.awaitAll()

                return@coroutineScope ArchitecturyVersion(versions.toSortedMap(), mcVersions)
            } catch (e: IOException) {
                e.printStackTrace()
                return@coroutineScope null
            }
        }
```

Earthcomputer (approved):

PR #1800, opened 2022-01-31
Use Paper repo for Velocity

NoahvdAa:
Velocity has also been moved over to the Paper repo.


NoahvdAa:
The old velocity repo will be shut down in four days (the 7th)

PR #1802, opened 2022-02-01
Add Velocity to readme.md

Astralchroma:
Noticed that it was missing, found it annoying so I fixed it.


Astralchroma:
Did anyone read this?

PR #1824, opened 2022-04-12
Remove multi module creators

Earthcomputer:
Multi-module creators were becoming a maintenance burden with little benefit. Multi-module projects are still supported, only the creators are being removed (so you have to create them manually from now on - most complex projects have significant customization in their buildscripts anyway). This will simplify adding new project creators in the future.


DenWav (approved):
Just did a quick glance, but it all looks good, and you know what you're doing.

Issue #1835, opened 2022-05-22
Add support for a single author to be listed in the plugin.yml


ShreyasAyyengar:
This is a very simple QOL feature that I would definitely find useful, usually as I am the only one writing my plugins. 

When creating a new project, and filling out the plugin.yml contents (eg. Main Class, Name, Dependencies) there is also a textbox for authors: 

If I wanted to put my name: `Shreyas Ayyengar`, the IntelliJ plugin would not let me proceed, stating that in order to list more than one author a comma (,) should split the names.  This means that those who want to list their name (with a space) as the only author, would not be able to do so.

Quite simple ¯\_(ツ)_/¯

syorito-hatsuki:
![image](https://user-images.githubusercontent.com/33298273/178137171-656ce35b-4a3f-46fb-a01e-eb1893e17fde.png)
Yep. Same trouble up to year

ShreyasAyyengar:
Any potential implementations? A short-term solution was provided in #1868 but was closed.

PR #1868, opened 2022-07-04
Allow spaces in listPattern

MSK1582:
This modifies the RegEx so that spaces can be placed in lists fixing #1835


DenWav:
Thanks for the PR. Ideally this would be more robust rather than just accepting a comma-separated list, but that was the easy option at the time. I don't think I want to merge this as-is, but this gives me an idea of how I want to handle this in the short term.

PR #1870, opened 2022-07-18
Fix Forge platform version detection.

marchermans:
The MinecraftDev plugin tries to detect the minecraft version from the name of the minecraft dependency.
However in doing so it assumed that the Forge Version would always be present, which is not the case if FG is run in vanilla mode.

In that vanilla mode the version string has a different format where the "_" is after the "-", causing an exception when the version substring is about the be loaded.

This fix checks for this additional condition and prevents the crash from occuring.


Mysterious-Dev (approved):

RedNesto (approved):
Sorry, I meant to review this earlier this week.

PR #1876, opened 2022-08-06
Support loom 0.13 and fix decompile options when using split Minecraft jars.

modmuss50:
Previously MCDev used internal APIs to get a number of values from loom. This PR moves those to use 2 new stable APIs I propose to add to loom. Backwards compatbility has been kept to ensure this still works with the current stable loom version.

Newer versions of loom support splitting the minecraft jar two common and client jars, this PR also ensures that the correct decompile task is used.

Loom PR adding new APIs: https://github.com/FabricMC/fabric-loom/pull/697

This is my first time working on an Idea plugin, and I have little Kotlin experience so please let me know if there is anything I can improve 👍

### DIFF
--- a/src/main/kotlin/platform/mcp/fabricloom/FabricLoomDecompileSourceProvider.kt
+++ b/src/main/kotlin/platform/mcp/fabricloom/FabricLoomDecompileSourceProvider.kt
@@ -10,7 +10,8 @@
 
 package com.demonwav.mcdev.platform.mcp.fabricloom
 
-import com.demonwav.mcdev.platform.fabric.util.FabricConstants
+import com.demonwav.mcdev.platform.forge.inspections.sideonly.Side
+import com.demonwav.mcdev.platform.forge.inspections.sideonly.SideOnlyUtil
 import com.demonwav.mcdev.util.findModule
 import com.demonwav.mcdev.util.runGradleTaskWithCallback
 import com.intellij.codeInsight.AttachSourcesProvider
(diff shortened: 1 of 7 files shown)


src/main/kotlin/platform/mcp/fabricloom/FabricLoomDecompileSourceProvider.kt
```diff
@@ -36,7 +37,23 @@ class FabricLoomDecompileSourceProvider : AttachSourcesProvider {
...
+
+        val decompileTasks = loomData.decompileTasks[env] ?: return emptyList()
+        return decompileTasks.map(::DecompileAction)
+    }
+
+    private fun isClientClass(psiFile: PsiJavaFile): Boolean {
+        return psiFile.classes.any { clazz ->
+            val environment = clazz.getAnnotation(FabricConstants.ENVIRONMENT_ANNOTATION)
+            val value = environment?.findDeclaredAttributeValue("value")?.text
+            return value == FabricConstants.ENV_TYPE_CLIENT
```
Earthcomputer: This doesn't work if someone puts spaces or other weird stuff in their annotations. It's also inefficient because it has to resolve the AST, which possibly involves loading and parsing the Java file if the annotation is a PSI stub. Normally you resolve the reference and check if the `PsiField` is equal to do stuff like this. However in this case, I'm fairly sure I have already implemented something in `SideOnlyUtil`?

Earthcomputer (changes requested):

src/main/kotlin/platform/mcp/fabricloom/FabricLoomDecompileSourceProvider.kt
modmuss50: All of these annotations here will have been auto generated by loom when spliting the mc jar so shouldnt be weird I hope. Ill take a look at SideOnlyUtil to see if it has what I need.

I had hoped to be able to read an entry from the jar's manifest but struggled to get that to work.

Earthcomputer (approved):

RedNesto (approved):

PR #1899, opened 2022-10-09
Access widener improvements

l-Luna:
- Implements go-to reference for descriptor parts
- Makes go-to-reference for method entries take into account method descriptors if present
- Adds an action for copying access widener entries for definitions, similar to the access transformer action
- Offers the action as a quick fix for inaccessible definitions (in Fabric projects only)
- Adds an inspection for duplicate access widener entries (same type + target)
- By improving shared code, this also improves the placement of balloons for other copy actions

Like other inspections, the duplicate inspection can be disabled globally. I tried to implement comment-based suppression, but changes to only comments didn't seem to apply new highlighting.

image
Untitled
Untitled2
Untitled3


src/main/kotlin/platform/mcp/aw/fixes/CopyAwAccessibleEntryFix.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.intellij.psi.PsiJavaCodeReferenceElement
+
+class CopyAwAccessibleEntryFix(val target: PsiElement, val element: PsiElement) : IntentionAction {
+
+    class Provider : UnresolvedReferenceQuickFixProvider<PsiJavaCodeReferenceElement>() {
+
+        override fun registerFixes(ref: PsiJavaCodeReferenceElement, registrar: QuickFixActionRegistrar) {
+            val module = ref.findModule() ?: return
+            val isFabric = MinecraftFacet.getInstance(module, FabricModuleType) != null
+            if (!isFabric) {
```
RedNesto, line 33: I'm not sure about this, other platforms might use access wideners without using Fabric (Sponge for example)

src/main/grammars/AwParser.bnf
```diff
@@ -88,20 +88,15 @@ member_name ::= NAME_ELEMENT {
...
-method_desc ::= OPEN_PAREN argument* CLOSE_PAREN return_value
+method_desc ::= OPEN_PAREN desc_element* CLOSE_PAREN desc_element
 
-field_desc ::= argument
+field_desc ::= desc_element
 
-argument ::= PRIMITIVE | CLASS_VALUE {
+desc_element ::= PRIMITIVE | CLASS_VALUE {
+	mixin= "com.demonwav.mcdev.platform.mcp.aw.psi.mixins.impl.AwDescElementImplMixin"
+	implements= "com.demonwav.mcdev.platform.mcp.aw.psi.mixins.AwDescElementMixin"
```
RedNesto: Can you replace those tabs with spaces?

src/main/kotlin/platform/mcp/aw/fixes/CopyAwAccessibleEntryFix.kt
RedNesto: I'm not sure about this, other platforms might use AWs outside of Fabric, Sponge for example.

I think the best solution would be to register the fix is an AW file exists in the module, but including some more platforms is enough to keep it simple.

src/main/kotlin/platform/mcp/aw/fixes/CopyAwAccessibleEntryFix.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.intellij.codeInsight.quickfix.UnresolvedReferenceQuickFixProvider
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.PsiJavaCodeReferenceElement
+
+class CopyAwAccessibleEntryFix(val target: PsiElement, val element: PsiElement) : IntentionAction {
+
+    class Provider : UnresolvedReferenceQuickFixProvider<PsiJavaCodeReferenceElement>() {
```
RedNesto, line 29: I wonder if this works with other JVM languages, or if we can use UAST here.

RedNesto (changes requested):
Looks good, just a few comments about the unresolved reference quickfix

src/main/kotlin/platform/mcp/aw/fixes/CopyAwAccessibleEntryFix.kt
l-Luna: These changes don't require tracking a project-wide access widener, I left that for a future PR. What other platforms use AWs other than Sponge?

src/main/kotlin/platform/mcp/aw/fixes/CopyAwAccessibleEntryFix.kt
l-Luna: I'll try it

src/main/kotlin/platform/mcp/aw/fixes/CopyAwAccessibleEntryFix.kt
RedNesto: I _think_ everything can use AWs if loaded early enough (similar to mixins), but spigot/bungee/velocity plugins are less likely to use them than on fabric/sponge (and forge ?)

src/main/kotlin/platform/mcp/aw/fixes/CopyAwAccessibleEntryFix.kt
l-Luna: Reimplemented in UAST and tested in Kotlin, Scala, and Groovy; it works in none of them.
Quick fixes for invisible elements as unresolved elements only exist in Java; other languages all detect this differently and don't have ways of providing fixes, let alone a unified interface. I'm not sure if this can be provided for Kotlin at all without implementing visibility checks ourselves.

src/main/kotlin/platform/mcp/aw/fixes/CopyAwAccessibleEntryFix.kt
l-Luna: So just Sponge for now?

RedNesto: Yeah it will be fine for now, thanks :)

src/main/kotlin/platform/mcp/aw/fixes/CopyAwAccessibleEntryFix.kt
RedNesto: Ah that's a shame, thanks for taking the time to try.

RedNesto (approved):

Mysterious-Dev (approved):

RedNesto:
Cancelling the macos run because it has been running for 2 hours...

PR #1918, opened 2022-12-28
Add Architectury to plugin description

Mysterious-Dev:
Everything is in the title.


Mysterious-Dev:
I hesitate to put the list in the same order as the project creator, tell me if it's better or not.

image

RedNesto (approved):
Architectury appearing after Forge and Fabric looks logical to me 👍

PR #1927, opened 2023-01-09
options.release change

Hephaisto-dev:
Hi, i wanted to change "options.release = value" with options.release.set(value) as release is a Property<Integer>. I made that to avoid warning when creating new projects.


DenWav:
I thought `=` was the correct method in Groovy?

Hephaisto-dev:
I don't know but i have a warning using Intellij

PR #1936, opened 2023-02-07
Fix FG2 gradle model not being instantiated

eigenraven:
In `src/main/kotlin/platform/mcp/gradle/datahandler/McpModelFG2Handler.kt` line 30, `resolverCtx.getExtraProject(gradleModule, McpModelFG2::class.java)` is used to obtain the FG2 Mcp model from the gradle plugin. However, the gradle plugin for FG2 did not implement `McpModelFG2` but only the superinterface `McpModel` which made that handler never activate.


Earthcomputer (approved):

Earthcomputer:
Just so you're aware, we're unlikely to put much more effort into supporting FG2 going forward, given how old it is. But I don't see a problem with simple PRs like this.

PR #1993, opened 2023-04-04
Make Gradle 7.6.1 Default Gradle

JT122406:
Current Default is 7.3.3 Which at this point is outdated by Gradle 7 Standards and doesn't support Java 18 and 19.  Gradle 7.6.1 is the latest with 7.6 being the last major 7.X version being released.


Mysterious-Dev (approved):
LGTM ^^.

PR #2024, opened 2023-05-02
Resolve deprecation warning on architectury

dayo05:


dayo05:
And, I suggest to increase default heap size of gradle too. Sometimes it breaks due to not enough ram space

RedNesto (approved):

PR #2029, opened 2023-05-07
Remap lines lazily in mixin targets that have not been decompiled yet

### LINKED ISSUE #2026: Incorrect line numbers when debugging decompiled class files that have mixins

2023.1-1.6.4

### IntelliJ version

2023.1.1 (#IC-231.8770.65)

### Operating System

Windows 10

### Target platform

Mixins

### Description of the bug

Classes that don't have sources attached and have mixins applied to them (or are mixins themselves) have incorrect line numbers in the debugger.

A simple way to reproduce this:
1. Import [fabric-example-mod](https://github.com/FabricMC/fabric-example-mod) (I used 1.18)
2. Place a breakpoint somewhere in `TitleScreen#init`
3. Run the game
4. See weirdness like this (I only placed a breakpoint on line 140)
idea64_2023-05-03_16-27-38

The issue is `MixinPositionManager` is emitting incorrect line numbers for decompiled classes. The solution seems to be calling `DebuggerUtilsEx.bytecodeToSourceLine` (like the default position manager, `PositionManagerImpl`, [does.](https://github.com/JetBrains/intellij-community/blob/6ff72f5ea5854027e77ab27d1020755e1f660bbf/java/debugger/impl/src/com/intellij/debugger/engine/PositionManagerImpl.java#L626-L633)) However, the class may not yet be decompiled at the time we are called, so a `RemappedSourcePosition` has to be returned in that case (again, see [`PositionManagerImpl`](https://github.com/JetBrains/intellij-community/blob/6ff72f5ea5854027e77ab27d1020755e1f660bbf/java/debugger/impl/src/com/intellij/debugger/engine/PositionManagerImpl.java#L158)).

makamys:
Fixes https://github.com/minecraft-dev/MinecraftDev/issues/2026#issuecomment-1537493401. Well, mostly.

Prior to this commit, switching to a stack frame inside a not yet decompiled class would take you to the wrong line, and switching to the same stack frame would always take you to the same wrong line until you restart the debug session.

With this commit, the first time you switch, you will still be taken to the wrong line, but on subsequent switches you will be taken to the correct line. The same thing happens even without the plugin, so I assume this is just a limitation of IntelliJ that can't be helped.

### DIFF
--- a/src/main/kotlin/platform/mixin/debug/MixinPositionManager.kt
+++ b/src/main/kotlin/platform/mixin/debug/MixinPositionManager.kt
@@ -29,6 +29,7 @@ import com.intellij.debugger.NoDataException
 import com.intellij.debugger.SourcePosition
 import com.intellij.debugger.engine.DebugProcess
 import com.intellij.debugger.engine.DebuggerUtils
+import com.intellij.debugger.engine.PositionManagerImpl.ClsSourcePosition
 import com.intellij.debugger.impl.DebuggerUtilsEx
 import com.intellij.debugger.requests.ClassPrepareRequestor
 import com.intellij.ide.highlighter.JavaFileType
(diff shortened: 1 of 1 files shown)


RedNesto:
Looks good :+1: Please fix the ktlint error and I'll merge it

makamys:
Whoops, fixed it.

RedNesto (approved):

PR #2090, opened 2023-07-25
Add inspection for when @Inject could be @Overwrite

### LINKED ISSUE #1836: Warn about unconditional inject-at-head-cancel mixins
This can be made with an Injector mixin that always cancels the rest of the function (at head) and possibly returns a different value.

This is effectively an overwrite with undefined behavior.
1. The ordering is undefined unless you change mixin priority. Two of these could conflict
2. Any transformations afterwards are entirely ignored. An overwrite would be more explicit about this behavior.

If an inject-at-head injector has a conditional that can change (unfamiliar with control flow in PSI, may not be feasible), then said warning should not be risen.

Earthcomputer:
Closes #1836

### DIFF
--- a/src/main/kotlin/platform/mixin/handlers/injectionPoint/ReturnInjectionPoint.kt
+++ b/src/main/kotlin/platform/mixin/handlers/injectionPoint/ReturnInjectionPoint.kt
@@ -21,7 +21,7 @@
 package com.demonwav.mcdev.platform.mixin.handlers.injectionPoint
 
 import com.demonwav.mcdev.platform.mixin.reference.MixinSelector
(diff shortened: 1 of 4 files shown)


LlamaLad7:
Which is best depends on the context. There are absolutely scenarios where you want to silently ignore all other changes to the method. Say you want to hide the hotbar unconditionally, is it really worth crashing when any mod tries to modify the rendering? No, they should have their stuff skipped silently. I don't think blindly forcing people to use `@Overwrite` is the right approach.

Earthcomputer:
Tbh I agree, I was just under the impression that this is what everyone wanted. I guess not, which makes me happy

Earthcomputer:
Gonna make the inspection off by default in that case

PR #2122, opened 2023-09-04
Mark paper-plugin.yml as experimental

lynxplay:
While paper plugins and their respective loading strategy have been integrated into paper for a while now, the developer facing side of paper plugins is still highly experimental.

Specifically the paper-plugin.yml file structure is still under active development and its syntax might change without a deprecation cycle. Its usage for general purpose plugin development is still hence discouraged.

While developers that are familiar with the caveats of using paper plugins but decide to do so anyway due to the already existing features exist, this commit still adds a warning to ensure that developers not familiar with the system do not accidentally use it for new projects, preventing frustration down the line with potential missing features or breaking changes over the usage of a plain plugin.yml file.


lynxplay:
Resolved formatting issue.

src/main/kotlin/platform/bukkit/creator/bukkit-platforms.kt
```diff
@@ -75,6 +77,10 @@ class PaperPlatformStep(parent: BukkitPlatformStep) : AbstractBukkitPlatformStep
         super.setupUI(builder)
         with(builder) {
             row("Paper manifest:") {
```
Leguan16:
```suggestion
            row("Paper Manifest:") {
```

lynxplay: Not my change in the PR, I don't think this is worth changing (at least on my side).

Leguan16: Its for consistency. Everything else is uppercase. Only manifest is lowercase. And since you currently touch this area I figured you can push a little commit to change it.

lynxplay: I don't have enough knowledge about the rest of the UI to tell you if this is actually convention or not so I don't feel comfortable with making the change. If we have a project maintainer to confirm this, that would be great. Otherwise I'll see if I find time to go through the rest of the UI to check.

RedNesto: This is correct, the other labels of the creator use title case, so feel free to apply it if you want to. I can do it easily later anyway.

lynxplay: Okay 👍 I'll switch it, thanks for the confirmation :)

PR #2127, opened 2023-09-21
MixinExtras support

LlamaLad7:
Adds full support for MixinExtras injectors and fixes some general functionality in related mixin areas.
See individual commits.

### DIFF
--- a/src/main/kotlin/platform/mixin/completion/MixinCompletionConfidence.kt
+++ b/src/main/kotlin/platform/mixin/completion/MixinCompletionConfidence.kt
@@ -30,7 +30,6 @@ import com.intellij.patterns.StandardPatterns
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
 import com.intellij.util.ThreeState
-import org.jetbrains.kotlin.idea.completion.or
 
 class MixinCompletionConfidence : CompletionConfidence() {
 
@@ -38,11 +37,11 @@ class MixinCompletionConfidence : CompletionConfidence() {
         .inside(
             false,
             PsiJavaPatterns.psiAnnotation().qName(
-                StandardPatterns.string().startsWith(MixinConstants.PACKAGE)
-                    .or(
-                        StandardPatterns.string()
-                            .oneOf(MixinAnnotationHandler.getBuiltinHandlers().map { it.first }.toList())
-                    )
+                StandardPatterns.or(
+                    StandardPatterns.string().startsWith(MixinConstants.PACKAGE),
+                    StandardPatterns.string()
+                        .oneOf(MixinAnnotationHandler.getBuiltinHandlers().map { it.first }.toList()),
+                )
             ),
             PlatformPatterns.psiFile(),
         )!!
--- a/src/main/kotlin/platform/mixin/completion/MixinCompletionConfidence.kt
+++ b/src/main/kotlin/platform/mixin/completion/MixinCompletionConfidence.kt
@@ -20,6 +20,7 @@
 
 package com.demonwav.mcdev.platform.mixin.completion
 
+import com.demonwav.mcdev.platform.mixin.handlers.MixinAnnotationHandler
 import com.demonwav.mcdev.platform.mixin.util.MixinConstants
 import com.intellij.codeInsight.completion.CompletionConfidence
 import com.intellij.codeInsight.completion.SkipAutopopupInStrings
@@ -29,13 +30,20 @@ import com.intellij.patterns.StandardPatterns
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
 import com.intellij.util.ThreeState
+import org.jetbrains.kotlin.idea.completion.or
 
 class MixinCompletionConfidence : CompletionConfidence() {
 
     private val mixinAnnotation = PlatformPatterns.psiElement()
         .inside(
             false,
-            PsiJavaPatterns.psiAnnotation().qName(StandardPatterns.string().startsWith(MixinConstants.PACKAGE)),
+            PsiJavaPatterns.psiAnnotation().qName(
+                StandardPatterns.string().startsWith(MixinConstants.PACKAGE)
+                    .or(
+                        StandardPatterns.string()
+                            .oneOf(MixinAnnotationHandler.getBuiltinHandlers().map { it.first }.toList())
+                    )
+            ),
             PlatformPatterns.psiFile(),
         )!!
 
--- a/src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt
+++ b/src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt
@@ -23,7 +23,6 @@ package com.demonwav.mcdev.platform.mixin.handlers.injectionPoint
 import com.demonwav.mcdev.platform.mixin.reference.isMiscDynamicSelector
 import com.demonwav.mcdev.platform.mixin.reference.parseMixinSelector
 import com.demonwav.mcdev.platform.mixin.reference.target.TargetReference
-import com.demonwav.mcdev.platform.mixin.util.MixinConstants.Annotations.CONSTANT
 import com.demonwav.mcdev.platform.mixin.util.MixinConstants.Annotations.SLICE
 import com.demonwav.mcdev.platform.mixin.util.findSourceElement
 import com.demonwav.mcdev.util.computeStringArray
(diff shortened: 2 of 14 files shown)


src/main/kotlin/platform/mixin/completion/MixinCompletionConfidence.kt
```diff
@@ -29,13 +30,20 @@ import com.intellij.patterns.StandardPatterns
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
 import com.intellij.util.ThreeState
+import org.jetbrains.kotlin.idea.completion.or
```
Earthcomputer: This adds a required dependency on the Kotlin plugin. Please find an alternative.

src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt
```diff
@@ -68,8 +74,13 @@ class AtResolver(
     private val targetMethod: MethodNode,
 ) {
     companion object {
+        private val injectionPointAnnotations = mapOf(
+            CONSTANT to "CONSTANT"
+        )
```
Earthcomputer: I'm already thinking this could be an extension point

src/main/kotlin/platform/mixin/handlers/mixinextras/MixinExtrasInjectorAnnotationHandler.kt
```diff
@@ -0,0 +1,338 @@
...
+import org.objectweb.asm.tree.AbstractInsnNode
+import org.objectweb.asm.tree.ClassNode
+import org.objectweb.asm.tree.FieldInsnNode
+import org.objectweb.asm.tree.LdcInsnNode
+import org.objectweb.asm.tree.MethodInsnNode
+import org.objectweb.asm.tree.MethodNode
+import org.objectweb.asm.tree.TypeInsnNode
+
+abstract class MixinExtrasInjectorAnnotationHandler : InjectorAnnotationHandler() {
+    open val oldSuperBehaviour = false
```
Earthcomputer: American English

src/main/kotlin/platform/mixin/handlers/mixinextras/MixinExtrasInjectorAnnotationHandler.kt
```diff
@@ -0,0 +1,338 @@
...
+
+    enum class ElementType {
```
Earthcomputer: Is this called an `ElementType` inside of MixinExtras? If not, maybe `InstructionType` would be a better name? `ElementType` could be confused with the type of a `PsiElement`.

src/main/kotlin/platform/mixin/handlers/mixinextras/MixinExtrasInjectorAnnotationHandler.kt
```diff
@@ -0,0 +1,338 @@
...
+                is Long -> return Type.LONG_TYPE
+                is Double -> return Type.DOUBLE_TYPE
+                is String -> return Type.getType(String::class.java)
+                is Type -> return Type.getType(Class::class.java)
+                else -> null
+            }
+        }
+
+        is TypeInsnNode -> {
+            return if (insn.getOpcode() < Opcodes.CHECKCAST) {
```
Earthcomputer, line 328: Super pedantic, but here I'd be worried that in the future they will add new instructions >`CHECKCAST` that are a `TypeInsnNode`, that Mixin doesn't treat as constants.

src/main/kotlin/platform/mixin/handlers/mixinextras/ModifyExpressionValueHandler.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.objectweb.asm.tree.ClassNode
+import org.objectweb.asm.tree.MethodNode
+
+class ModifyExpressionValueHandler : MixinExtrasInjectorAnnotationHandler() {
+    override val supportedElementTypes = listOf(
+        ElementType.METHOD_CALL, ElementType.FIELD_GET, ElementType.INSTANTIATION, ElementType.CONSTANT
+    )
+
+    override fun extraTargetRestrictions(insn: AbstractInsnNode) =
+        getInsnReturnType(insn)?.equals(Type.VOID_TYPE) == false
```
Earthcomputer: Avoid using `equals` in Kotlin code:
```suggestion
    override fun extraTargetRestrictions(insn: AbstractInsnNode): Boolean {
        val returnType = getInsnReturnType(insn) ?: return false
        return returnType != Type.VOID_TYPE
    }
```
Note that the simpler `getInsnReturnType(insn) != Type.VOID_TYPE` is not equivalent to what you've written, because it returns `true` if `getInsnReturnType(insn)` is `null`.

src/main/kotlin/platform/mixin/handlers/mixinextras/WrapOperationHandler.kt
```diff
@@ -0,0 +1,72 @@
...
+import com.intellij.psi.PsiAnnotation
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiPrimitiveType
+import com.intellij.psi.PsiType
+import com.intellij.psi.search.GlobalSearchScope
+import org.objectweb.asm.tree.AbstractInsnNode
+import org.objectweb.asm.tree.ClassNode
+import org.objectweb.asm.tree.MethodNode
+
+private const val OPERATION = "com.llamalad7.mixinextras.injector.wrapoperation.Operation"
```
Earthcomputer: This should go in `MixinConstants`

Earthcomputer (changes requested):
Thank you! Just some things

src/main/kotlin/platform/mixin/handlers/mixinextras/MixinExtrasInjectorAnnotationHandler.kt
LlamaLad7: This whole method is copied from Mixin including that check

src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt
LlamaLad7: It wouldn't really have an implementation class? I agree it should probably be extensible but I'm not sure exactly how to approach that

src/main/kotlin/platform/mixin/handlers/mixinextras/ModifyExpressionValueHandler.kt
LlamaLad7: That's why I did it this way in the first place. If your version is deemed to be cleaner then I can change it.

src/main/kotlin/platform/mixin/handlers/mixinextras/MixinExtrasInjectorAnnotationHandler.kt
Earthcomputer: Please change this field to use American English. In this case, this would be `oldSuperBehavior`, rather than `oldSuperBehaviour`.

src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt
Earthcomputer: Extension points don't need an implementation class. You can create a bean class which contains the annotation name and the `atCode`, and specify the extension point and the bean class in the `plugin.xml`. Your EP name would be a `ExtensionPointName<YourBeanClass>`, which is essentially a list of `YourBeanClass`. If you're worried about performance, you can create a map from that and update it dynamically using an extension point listener.

Earthcomputer: ~~Also maybe rename this to `impliedInjectionPoints` or something along those lines.~~

Nvm, I briefly misunderstood what this map was for. I do think if we're making it an extension point we should put more thought into the name though.

src/main/kotlin/platform/mixin/handlers/mixinextras/MixinExtrasInjectorAnnotationHandler.kt
Earthcomputer: Ah, whatever then. If new instructions do emerge and Mixin does get broken, we will at least be consistent with Mixin's broken behavior.

src/main/kotlin/platform/mixin/handlers/mixinextras/MixinExtrasInjectorAnnotationHandler.kt
LlamaLad7: That was meant to be more of a "smh", didn't realise what the emoji was called \:P

src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt
LlamaLad7: Ah, gotcha. I misunderstood the point of the bean class.

LlamaLad7: Something like `annotationAliasForInjectionPoint` maybe? It's kinda hard to explain succinctly

Earthcomputer: On second thought, I think `injectionPointAnnotation` is fine.

Earthcomputer (approved):
Tyvm!

PR #2130, opened 2023-09-23
Add external annotations to Minecraft libraries

Earthcomputer:
This is to support a switch to using a `@Translatable` annotation for translation inspections. See https://github.com/minecraft-dev/annotations/pull/1.

Note that this is not to support general addition or removal of annotations to the Minecraft code (such as removing `@Deprecated` from the `Block` class), only those to support MinecraftDev. General addition or removal is in my opinion better suited in other projects such as Parchment and Yarn.

### DIFF
--- a/src/main/kotlin/translations/identification/TranslationAnnotationsLocationProvider.kt
+++ b/src/main/kotlin/translations/identification/TranslationAnnotationsLocationProvider.kt
@@ -1,0 +1,48 @@
+/*
+ * Minecraft Development for IntelliJ
+ *
+ * https://mcdev.io/
+ *
+ * Copyright (C) 2023 minecraft-dev
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, version 3.0 only.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.demonwav.mcdev.translations.identification
+
+import com.demonwav.mcdev.platform.mcp.framework.MCP_LIBRARY_KIND
+import com.intellij.codeInsight.externalAnnotation.location.AnnotationsLocation
+import com.intellij.codeInsight.externalAnnotation.location.AnnotationsLocationProvider
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.roots.OrderRootType
+import com.intellij.openapi.roots.libraries.Library
+import com.intellij.openapi.roots.ui.configuration.libraries.LibraryPresentationManager
+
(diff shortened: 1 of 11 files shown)


src/main/kotlin/translations/identification/TranslationAnnotationsLocationProvider.kt
```diff
@@ -0,0 +1,48 @@
...
+
+import com.demonwav.mcdev.platform.mcp.framework.MCP_LIBRARY_KIND
+import com.intellij.codeInsight.externalAnnotation.location.AnnotationsLocation
+import com.intellij.codeInsight.externalAnnotation.location.AnnotationsLocationProvider
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.roots.OrderRootType
+import com.intellij.openapi.roots.libraries.Library
+import com.intellij.openapi.roots.ui.configuration.libraries.LibraryPresentationManager
+
+class TranslationAnnotationsLocationProvider : AnnotationsLocationProvider {
```
RedNesto, line 31: I wonder if and how this and the artifact resolver could be generalized for potential future annotations.
Not a request just a thought.

src/main/kotlin/translations/identification/TranslationExternalAnnotationsArtifactsResolver.kt
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        return runWriteAction {
+            val annotationsPath = findAnnotationsPath(false)
+                ?: findAnnotationsPath(true)
+                ?: return@runWriteAction false
+
+            val editor = ExistingLibraryEditor(library, null)
+            val type = AnnotationOrderRootType.getInstance()
+            editor.getUrls(type).forEach { editor.removeRoot(it, type) }
```
RedNesto: Won't this remove existing external annotations ?
Could there be a way to only add those provided by the plugin ?

src/main/kotlin/translations/identification/TranslationExternalAnnotationsArtifactsResolver.kt
```diff
@@ -0,0 +1,130 @@
...
+
+        return promise
+    }
+
+    private fun doResolve(library: Library): Boolean {
+        if (library !is LibraryEx || library.isDisposed) {
+            return false
+        }
+
+        return runWriteAction {
```
RedNesto, line 85: Make sure to check that the library is not disposed inside the write action.

RedNesto (changes requested):

src/main/kotlin/translations/identification/TranslationExternalAnnotationsArtifactsResolver.kt
Earthcomputer: I copied from the `ExternalAnnotationsRepositoryResolver` here, it seems to work fine without it though so I'll remove it.

src/main/kotlin/translations/identification/TranslationExternalAnnotationsArtifactsResolver.kt
Earthcomputer: I copied from `ExternalAnnotationsRepositoryResolver` again here, I don't think it's possible for the library to be disposed between these two points because `runWriteAction` is sync and we're on the EDT

RedNesto: Right, sounds good then 👍

RedNesto (approved):

PR #2133, opened 2023-09-24
Suppress `remap = true` warning in `@At`s

LlamaLad7:
If the enclosing injector annotation has a `remap = false` then by default this value is inherited so an explicit `remap = true` is necessary.


Earthcomputer (changes requested):
This feature already exists, if it doesn't work then that's a bug. Could you take a look at `DefaultAnnotationParamInspectionSuppressor` to see why?

LlamaLad7:
Superseded by #2134

PR #2136, opened 2023-09-25
Fix WrongOperationParametersInspection

LlamaLad7:
New IDEA versions seemingly compare the text differently, including the generic. This is a more robust check anyway.


Earthcomputer:
https://github.com/minecraft-dev/MinecraftDev/commit/2e509b4344006cc9877f72b991573cdfbf0c55da

LlamaLad7:
I need to start looking more carefully before doing stuff xD
Thanks for the fix!

PR #2216, opened 2024-01-29
Simplified Chinese translation

TexBlock:


Earthcomputer (approved):
Just be aware that mcdev is only half translatable at the moment, there may be more translations for existing things added in the future.

PR #2244, opened 2024-02-25
MixinExtras: Add handler signature support for expressions.

LlamaLad7:
There are still 2 things to address, but you'd probably be better placed to implement them than me.
1. It doesn't handle `int-like` types. Ideally the user would be presented with a dropdown of the suitable types so they can choose. All int-like types within the handler should match (at least with the current possible targets). If that is too much to ask then it should probably just treat them as `int` types, but I do worry the users would get confused when they expected e.g. a char and have to change it themselves.
2. Not a direct result of this PR but the name suggestion code suggests duplicate names for parameters of the same type, which causes issues. This is noticeable e.g. when wrapping a comparison.

(I don't want this to target dev but I don't seem to have a choice. It can be manually merged. Review the latest commit only.)

### DIFF
--- a/src/main/kotlin/platform/mixin/handlers/mixinextras/TargetInsn.kt
+++ b/src/main/kotlin/platform/mixin/handlers/mixinextras/TargetInsn.kt
@@ -1,0 +1,30 @@
+/*
+ * Minecraft Development for IntelliJ
+ *
+ * https://mcdev.io/
+ *
+ * Copyright (C) 2024 minecraft-dev
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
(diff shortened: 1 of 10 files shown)


LlamaLad7:
...

RedNesto:
Erm I guess this should only trigger when targeting a release branch

src/main/kotlin/platform/mixin/handlers/mixinextras/TargetInsn.kt
```diff
@@ -0,0 +1,10 @@
+package com.demonwav.mcdev.platform.mixin.handlers.mixinextras
```
Earthcomputer, line 21: Please add the missing license header

Earthcomputer (changes requested):

PR #2255, opened 2024-03-06
Don't check `@Final` for fields with initializers

ctrlaltmilk:
Mixin `@Shadow` fields only require `@Final` when they don't have an initializer. When they do have an initializer, Mixin will throw a warning saying to just use the `final` modifier. This fixes the inspection to no longer apply to fields with an initializer.

### DIFF
--- a/src/main/kotlin/platform/mixin/inspection/shadow/ShadowModifiersInspection.kt
+++ b/src/main/kotlin/platform/mixin/inspection/shadow/ShadowModifiersInspection.kt
@@ -103,8 +103,10 @@ class ShadowModifiersInspection : MixinInspection() {
             // TODO: Would it make sense to apply the @Final check to methods?
             if (member !is PsiField) {
                 return
-            } else if (member.hasInitializer()) {
-                // @Final annotation doesn't apply to members that are initialized in the mixin class
+            }
+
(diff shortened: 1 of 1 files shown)


src/main/kotlin/platform/mixin/inspection/shadow/ShadowModifiersInspection.kt
```diff
@@ -103,6 +103,9 @@ class ShadowModifiersInspection : MixinInspection() {
             // TODO: Would it make sense to apply the @Final check to methods?
             if (member !is PsiField) {
                 return
+            } else if (member.hasInitializer()) {
```
Earthcomputer: Nit: could you change this from an `else if` to a regular `if`?

ctrlaltmilk: I'll check if type inference still holds in that situation - I want to avoid an ugly cast to `PsiField` and I figured this was the easiest way to do it.

Earthcomputer: It should work because of the return

Earthcomputer (approved):

PR #2274, opened 2024-03-31
Add support for MixinExtras expressions

Earthcomputer:
See [Llama's gist](https://gist.github.com/LlamaLad7/ac2f4ced03d21afddde80c5c5e9b302f) for a (slightly outdated) overview of what MixinExtras expressions are. Without IDE support, ME expressions would be really inconvenient to write, more so than other mixin features.

ME expression syntax is support through a custom language in IntelliJ, through the use of language injection. To make this nicer to work with, I have slightly extended the ME expression language. Definitions are prefixed with the `class` keyword, and then a boolean literal for whether they are a type or not. Expressions are surrounded with `do { ... }`. For example:
```java
@Definition(id = "Integer", type = Integer.class)
@Definition(id = "x", local = @Local(type = Object.class, ordinal = 0))
@Expression("(Integer) x")
```
would produce the following injected ME expression file:
```java
class true Integer
class false x
do { (Integer) x }
```

This feature is basically done now, the only thing left is to switch away from the JitPack version of MixinExtras and use the library they'll provide for us, which will hopefully have the same interface.

Don't worry about the amount of commits, I'll squash them all on merge.


src/main/resources/META-INF/plugin.xml
```diff
@@ -432,6 +438,7 @@
         <lang.foldingBuilder language="JAVA" implementationClass="com.demonwav.mcdev.platform.mixin.folding.MixinObjectCastFoldingBuilder"/>
         <lang.foldingBuilder language="JAVA" implementationClass="com.demonwav.mcdev.platform.mixin.folding.MixinTargetDescriptorFoldingBuilder"/>
         <lang.foldingBuilder language="JAVA" implementationClass="com.demonwav.mcdev.platform.mixin.folding.AccessorMixinFoldingBuilder"/>
+        <lang.foldingBuilder language="JAVA" implementationClass="com.demonwav.mcdev.platform.mixin.expression.MEDefinitionFoldingBuilder"/>
```
RedNesto, line 479: Nitpick: why is this not with the other ME EPs a bit further down?

src/main/kotlin/platform/mixin/util/LocalVariables.kt
```diff
@@ -855,6 +866,7 @@ object LocalVariables {
         val type: PsiType,
         val index: Int,
         val mixinName: String = name,
+        val variable: PsiVariable? = null,
```
RedNesto, line 873: I'm not super familiar with that part of the Mixin support, but having fields containing direct references to PsiElements might end up causing leaks, using pointers would be more appropriate.

src/main/kotlin/util/psi-utils.kt
```diff
@@ -226,6 +253,42 @@ inline fun <T> PsiElement.cached(vararg dependencies: Any, crossinline compute:
...
+
+    cacheLock.readLock().lock()
+    try {
+        val value = getUserData(key)?.upToDateOrNull
+        if (value != null) {
+            return value.get()
+        }
+    } finally {
+        cacheLock.readLock().unlock()
+    }
```
RedNesto: This is not super consequential, but using Kotlin's `withLock` extension function avoids that Java try/finally-unlock boilerplate.

build.gradle.kts
```diff
@@ -86,12 +86,25 @@ repositories {
...
 }
 
 dependencies {
     // Add tools.jar for the JDI API
     implementation(files(Jvm.current().toolsJar))
 
+    // TODO: temporary waiting for MixinExtras expression library
+    testLibs(implementation("com.github.LlamaLad7.MixinExtras:mixinextras-common:290f3d8")!!)
+    implementation("org.spongepowered:mixin:0.8.4")
+    implementation("org.ow2.asm:asm-util:9.3")
```
DenWav, line 119: Add the Maven coordinates to the `gradle/libs.versions.toml` file.

src/main/grammars/MEExpressionLexer.flex
```diff
@@ -0,0 +1,147 @@
...
+%%
+
+%public
+%class MEExpressionLexer
+%implements FlexLexer
+%unicode
+%function advance
+%type IElementType
+%eof{ return;
+%eof}
```
DenWav: What's this here for?

Referring to this:
```
%eof{ return;
%eof}
```

src/main/grammars/MEExpressionLexer.flex
```diff
@@ -0,0 +1,147 @@
...
+import com.intellij.psi.TokenType;
+
+%%
+
+%public
+%class MEExpressionLexer
+%implements FlexLexer
+%unicode
+%function advance
+%type IElementType
```
DenWav, line 34: Nit picking (I'm sorry, I know) but I like to just keep this bit consistent with the other flex files we have - like [this](https://github.com/minecraft-dev/MinecraftDev/blob/dev/src/main/grammars/AtLexer.flex#L36-L40), with `%unicode` after the states.

The only reason I point this out is just because I find it easier to keep track when they're the same.

src/main/grammars/MEExpressionLexer.flex
```diff
@@ -0,0 +1,147 @@
...
+
+STRING_TERMINATOR = '
+STRING_ESCAPE = \\'|\\\\
+
+%state STRING
+
+%%
+
+<YYINITIAL> {
+    {WHITE_SPACE}+ { yybegin(YYINITIAL); return TokenType.WHITE_SPACE; }
```
DenWav: It's not necessary to call `yybegin(YYINITIAL)` when you're already in the `YYINITIAL` state. You only need that if you're changing state.

src/main/kotlin/platform/mixin/expression/MEExpressionInjector.kt
```diff
@@ -0,0 +1,205 @@
...
+        val modifierList = anchor.findContainingModifierList() ?: return
+
+        val modCount = PsiModificationTracker.getInstance(project).modificationCount
+        val primaryElement = modifierList.getUserData(ME_EXPRESSION_INJECTION)
+        if (primaryElement != null && primaryElement.modCount == modCount) {
+            METHOD_ADD_TO_RESULTS.invoke(registrar, primaryElement.injectionResult)
+            return
+        }
+
+        var isFrankenstein = false
```
DenWav, line 99: Can you explain what frankenstein means with a comment here?

DenWav (changes requested):
Well this PR is insane. I don't really know anything about MixinExtras, and can barely follow parts of this PR, and can't follow other parts at all. The code seems fine, it's just going over my head in the more complicated places, but I don't have a problem with trusting you. I did look at (most) of the code in this PR at least, even if I don't really know all of what's going on.

I apologize that the only comments I have are unimportant nitpicks, again I'm going mostly on trust here. Get the tests passing and address these and I'm happy to call this one good to go.

RedNesto (commented):
This looks really good.

I don't have much to say about the features added as I'm not familiar with MixinExtras.

Also not sure about those `invokeLater { WriteCommandAction[...] }` patterns. It might be the only way to achieve this today, but it might age poorly because of https://youtrack.jetbrains.com/issue/IJPL-53

src/main/grammars/MEExpressionLexer.flex
Earthcomputer: No idea. I copied it from the IntelliJ language tutorial

build.gradle.kts
Earthcomputer: These are all temporary while we wait for the MixinExtras standalone library, when they'll be transitive dependencies.

src/main/grammars/MEExpressionLexer.flex
Earthcomputer: Really? I didn't realize

src/main/kotlin/platform/mixin/util/LocalVariables.kt
Earthcomputer: I don't think this object is supposed to be stored anywhere. I'll add that to the docs though.

Earthcomputer:
I think once write actions are allowed outside of the EDT we'll have to make a general sweep of the codebase to look for cases like these where we use `invokeLater` for write actions.

src/main/kotlin/util/psi-utils.kt
Earthcomputer: It didn't let me do this for some reason, prompting me to add kotlin stdlib to the classpath, but don't we already have stdlib-jdk8?

LlamaLad7: Perhaps more to the point kt also has `ReentrantReadWriteLock().read { }` and a similar write method.

LlamaLad7 (approved):

src/main/kotlin/util/psi-utils.kt
Earthcomputer: Couldn't find `ReentrantReadWriteLock.read`

src/main/resources/META-INF/plugin.xml
Earthcomputer: Because I put it in the folding category

src/main/kotlin/util/psi-utils.kt
LlamaLad7: It's only on `ReentrantReadWriteLock`, not `ReadWriteLock`, so the key's type needs to be more specific.

Earthcomputer: Ah yes sorry. But it's also in kotlin stdlib, and it's giving me the same problem as `withLock`.

LlamaLad7: Might be an IDEA cache issue? I can build fine having switched to `read` and `write`

Earthcomputer: Yeah must have been. I can use it now.

PR #2280, opened 2024-04-12
Suppress `ReferenceToMixin` for `@Dynamic`

### LINKED ISSUE #629: Reference to @Mixin class inspection applied to @Dynamic#mixin
```
Minecraft Development v2019.1-1.2.22
IntelliJ IDEA 2019.1.3 (Community Edition)
Build #IC-191.7479.19, built on May 28, 2019
Windows 10 10.0
```

The `Reference to @Mixin class` inspection is still applied to [`@Dynamic#mixin`](https://github.com/SpongePowered/Mixin/blob/master/src/main/java/org/spongepowered/asm/mixin/Dynamic.java#L105), which is explicitly designed for mixin class references.

inspection screenshot

senseiwells:
Fixes #629


Earthcomputer:
It's probably better for this logic to be added to `MixinClassReferenceInspection`, since we can change that code. Inspection suppressors tend to be a bit brittle and make the logic of why an inspection isn't appearing hard to follow. Generally they should only be used if it's logically an inspection suppression (e.g. `@SuppressWarnings`, `@Dynamic` when it behaves like a suppression) or if there is no alternative.

Earthcomputer (approved):

RedNesto (approved):

PR #2287, opened 2024-04-23
MixinSquared Support

Bawnorton:
### Support for [MixinSquared](https://github.com/Bawnorton/MixinSquared)
Not  familiar with this framework, or kotlin, so I may have re-invented a wheel somewhere.

Things that I believe can be improved but don't know the best way to go about it:
- Finding the registration calls seems inefficient 
https://github.com/Bawnorton/MinecraftDev/blob/5376600746fe5e7a57a6f50999065afd0de1dc64/src/main/kotlin/platform/mixin/reference/MixinSelectors.kt#L462
- Needed a way to have the target class be selector dependent rather than mixin dependent
https://github.com/Bawnorton/MinecraftDev/blob/5376600746fe5e7a57a6f50999065afd0de1dc64/src/main/kotlin/platform/mixin/reference/MixinSelectors.kt#L397
    results in patterns similar to this
https://github.com/Bawnorton/MinecraftDev/blob/5376600746fe5e7a57a6f50999065afd0de1dc64/src/main/kotlin/platform/mixin/handlers/InjectorAnnotationHandler.kt#L64-L75

Todo:
- [x] Misc dynamic selectors have their namespaces properly identified
- [x] `@TargetHandler` attributes navigate to the members they represent
- [x] `@TargetHandler` attributes have inspections for when their targets are not found
- [x] `@TargetHandler` prefixes map to their injectors with inspections for missing or invalid prefixes
- [x] Dynamic selector `MixinSquared:Handler` resolves to the identified target
- [x] `@TargetHandler` mixin attribute folds to the simple class name
- [ ] `@TargetHandler` attribute preview on hover
- [ ] `@TargetHandler` attribute auto-completion
- [ ] Inspection for mixin priority, if target mixin will apply after, error on target handler annotation
- [ ] Fix targeting for the first lambda method created in target mixin constructors
    - Not entirely sure why this breaks as the `name` parameter uses the `MemberReference` selector which can handle lambda references fine
    - Other lambdas are handled fine, it's specifically the first one found in a target mixin constructor that fails
- [ ] Tests


Earthcomputer:
Hi, do you have any metrics on how often MixinSquared is actually used? I'm happy to add this if it turns out it's used a lot, but otherwise I'm leaning towards this being a separate plugin to be installed alongside mcdev. Our Mixin support is designed to be extensible by other plugins for situations like this, but if you encounter any issues with it I'm happy to accept fixes. That said, if it does turn out to be popular enough (like MixinExtras turned out to be), it may be worth putting it in mcdev directly so users don't have to install another plugin.

Bawnorton:
> Hi, do you have any metrics on how often MixinSquared is actually used? I'm happy to add this if it turns out it's used a lot, but otherwise I'm leaning towards this being a separate plugin to be installed alongside mcdev. Our Mixin support is designed to be extensible by other plugins for situations like this, but if you encounter any issues with it I'm happy to accept fixes. That said, if it does turn out to be popular enough (like MixinExtras turned out to be), it may be worth putting it in mcdev directly so users don't have to install another plugin.

The main metric I use is GH search as I don't track anything myself:
https://github.com/search?q=com.bawnorton.mixinsquared+-user%3ABawnorton&type=code

Jitpack's download count for this month is currently at 1.5k:
https://jitpack.io/#Bawnorton/MixinSquared

It is decently used, definitely not nearly as much as mixinextras due to the use cases being rarer.

I have received requests from people who use mixinsquared to add support to mcdev so that was the motivation for this pr:
Screenshot 2024-04-23 at 5 48 02 PM

Separately
https://github.com/minecraft-dev/MinecraftDev/blob/1e828e178ab8423c8225d7d75ce2e6dd7b5ab574/src/main/kotlin/platform/mixin/reference/MixinSelectors.kt#L417-L427

does not account for dynamic selectors having their namespace specified at registration instead of in the annotation, thus, mcdev does not recognise the `@MixinSquared:Handler` as a valid target, but it does recognise `@Handler` as valid (which it actually isn't). The end result is you need to suppress 3 different inspections to remove the error highlighting.

Namespace at registration:
https://github.com/Bawnorton/MixinSquared/blob/a4209feab1fe64a26acf6d09e8b56e874f132847/src/main/java/com/bawnorton/mixinsquared/MixinSquaredBootstrap.java#L47

The fix for this is bundled into this pr but I am not sure if it is a good fix or if there is a better way. Reading the `TargetSelector#dynamicSelectors` field from the would be ideal but didn't seem possible from what I experimented with and what I found.

I did consider creating a separate plugin for this, however, some of the changes I needed to make to allow for the targeting to work required adapting various internals which are not currently exposed so I felt it wouldn't make much sense to try and force my way around the current structure from an external plugin that depends on mcdev.

Bawnorton:
I'm gonna move this to a separate plugin and pr any necessary changes I'll need.

Earthcomputer:
> I'm gonna move this to a separate plugin and pr any necessary changes I'll need.

That sounds reasonable. Thanks!

PR #2292, opened 2024-05-03
Translations: force json, convert template, default to existing key

Moulberry:
Adds 3 new translation features (I can split into 3 PRs if needed)

1. Add Force JSON Translation
For me (and maybe others), the plugin isn't able to detect the Minecraft version and throws "Cannot determine MC version". This option makes it avoid erroring

2. Add template for the Convert to Translation feature
net.minecraft.client.resources.I18n isn't always desired, this option allows the template string to be changed

3. Default to existing key when adding duplicate translation text
Scans the translation file for an existing key that matches the translation text and suggests it inside the Convert to Translation dialog

Feature 3 only supports json translation files, support for lang files can be added if desired.

Apologies if there are any stylistic issues, I am not a Kotlin developer

### DIFF
--- a/src/main/kotlin/MinecraftConfigurable.kt
+++ b/src/main/kotlin/MinecraftConfigurable.kt
@@ -33,10 +33,7 @@ import com.intellij.ui.dsl.builder.AlignX
 import com.intellij.ui.dsl.builder.BottomGap
 import com.intellij.ui.dsl.builder.bindItem
 import com.intellij.ui.dsl.builder.bindSelected
-import com.intellij.ui.dsl.builder.bindText
 import com.intellij.ui.dsl.builder.panel
-import com.intellij.ui.dsl.builder.selected
-import com.intellij.ui.layout.ComponentPredicate
 import com.intellij.util.IconUtil
 import javax.swing.JComponent
 import org.jetbrains.annotations.Nls
@@ -91,25 +88,6 @@ class MinecraftConfigurable : Configurable {
             row {
                 checkBox(MCDevBundle("minecraft.settings.mixin.shadow_annotation_same_line"))
                     .bindSelected(settings::isShadowAnnotationsSameLine)
-            }
-        }
-
-        group(MCDevBundle("minecraft.settings.translation")) {
-            row {
-                checkBox(MCDevBundle("minecraft.settings.translation.force_json_translation_file"))
-                    .bindSelected(settings::isForceJsonTranslationFile)
-            }
-
-            lateinit var allowConvertToTranslationTemplate: ComponentPredicate
-            row {
-                val checkBox = checkBox(MCDevBundle("minecraft.settings.translation.use_custom_convert_template"))
-                    .bindSelected(settings::isUseCustomConvertToTranslationTemplate)
-                allowConvertToTranslationTemplate = checkBox.selected
-            }
-
-            row {
-                textField().bindText(settings::convertToTranslationTemplate)
-                    .enabledIf(allowConvertToTranslationTemplate)
             }
         }
 
--- a/src/main/kotlin/MinecraftConfigurable.kt
+++ b/src/main/kotlin/MinecraftConfigurable.kt
@@ -29,7 +29,12 @@ import com.intellij.openapi.project.ProjectManager
 import com.intellij.openapi.ui.DialogPanel
 import com.intellij.ui.EnumComboBoxModel
 import com.intellij.ui.components.Label
-import com.intellij.ui.dsl.builder.*
+import com.intellij.ui.dsl.builder.AlignX
+import com.intellij.ui.dsl.builder.BottomGap
+import com.intellij.ui.dsl.builder.bindItem
+import com.intellij.ui.dsl.builder.bindSelected
+import com.intellij.ui.dsl.builder.bindText
+import com.intellij.ui.dsl.builder.panel
 import com.intellij.util.IconUtil
 import javax.swing.JComponent
 import org.jetbrains.annotations.Nls
@@ -95,7 +100,6 @@ class MinecraftConfigurable : Configurable {
             row(MCDevBundle("minecraft.settings.translation.convert_to_translation_template")) {
                 textField().bindText(settings::convertToTranslationTemplate)
             }
-
         }
 
         onApply {
--- a/src/main/kotlin/MinecraftConfigurable.kt
+++ b/src/main/kotlin/MinecraftConfigurable.kt
@@ -100,16 +100,16 @@ class MinecraftConfigurable : Configurable {
                     .bindSelected(settings::isForceJsonTranslationFile)
             }
 
-            lateinit var property: ComponentPredicate
+            lateinit var allowConvertToTranslationTemplate: ComponentPredicate
             row {
                 val checkBox = checkBox(MCDevBundle("minecraft.settings.translation.use_custom_convert_template"))
                     .bindSelected(settings::isUseCustomConvertToTranslationTemplate)
-                property = checkBox.selected
+                allowConvertToTranslationTemplate = checkBox.selected
             }
 
             row {
                 textField().bindText(settings::convertToTranslationTemplate)
-                    .enabledIf(property)
+                    .enabledIf(allowConvertToTranslationTemplate)
             }
         }
 
--- a/src/main/kotlin/MinecraftConfigurable.kt
+++ b/src/main/kotlin/MinecraftConfigurable.kt
@@ -29,11 +29,7 @@ import com.intellij.openapi.project.ProjectManager
 import com.intellij.openapi.ui.DialogPanel
 import com.intellij.ui.EnumComboBoxModel
 import com.intellij.ui.components.Label
-import com.intellij.ui.dsl.builder.AlignX
-import com.intellij.ui.dsl.builder.BottomGap
-import com.intellij.ui.dsl.builder.bindItem
-import com.intellij.ui.dsl.builder.bindSelected
-import com.intellij.ui.dsl.builder.panel
+import com.intellij.ui.dsl.builder.*
 import com.intellij.util.IconUtil
 import javax.swing.JComponent
 import org.jetbrains.annotations.Nls
@@ -89,6 +85,17 @@ class MinecraftConfigurable : Configurable {
                 checkBox(MCDevBundle("minecraft.settings.mixin.shadow_annotation_same_line"))
                     .bindSelected(settings::isShadowAnnotationsSameLine)
             }
+        }
+
+        group(MCDevBundle("minecraft.setting.translation")) {
+            row {
+                checkBox(MCDevBundle("minecraft.settings.translation.force_json_translation_file"))
+                    .bindSelected(settings::isForceJsonTranslationFile)
+            }
+            row(MCDevBundle("minecraft.settings.translation.convert_to_translation_template")) {
+                textField().bindText(settings::convertToTranslationTemplate)
+            }
+
         }
 
         onApply {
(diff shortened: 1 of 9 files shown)


src/main/kotlin/MinecraftSettings.kt
```diff
@@ -37,6 +37,9 @@ class MinecraftSettings : PersistentStateComponent<MinecraftSettings.State> {
         var underlineType: UnderlineType = UnderlineType.DOTTED,
 
         var isShadowAnnotationsSameLine: Boolean = true,
+
+        var isForceJsonTranslationFile: Boolean = false,
+        var convertToTranslationTemplate: String = "net.minecraft.client.resources.I18n.format(\"\$key\")"
```
LlamaLad7: Trailing comma

src/main/kotlin/translations/TranslationFiles.kt
```diff
@@ -223,6 +260,17 @@ object TranslationFiles {
...
     }
 
+    private fun PsiFile.findKeyForTextAsJson(text: String): String? {
+        val rootObject = this.firstChild as? JsonObject ?: return null
+        return rootObject.propertyList.firstOrNull {
+            if (it.value is JsonStringLiteral) {
+                (it.value as JsonStringLiteral).value == text
+            } else {
+                false
+            }
```
LlamaLad7: Can be `(it.value as? JsonStringLiteral)?.value == text`

src/main/kotlin/translations/intentions/ConvertToTranslationIntention.kt
```diff
@@ -73,12 +77,14 @@ class ConvertToTranslationIntention : PsiElementBaseIntentionAction() {
             val key = result.first ?: return
             val replaceLiteral = result.second
             try {
-                TranslationFiles.add(element, key, value)
+                if (existingKey == null || existingKey != key) {
```
LlamaLad7: No need for the null check, `==` is null-safe.

src/main/kotlin/translations/TranslationFiles.kt
```diff
@@ -110,12 +111,48 @@ object TranslationFiles {
...
+
+        for (file in files) {
+            val psiFile = PsiManager.getInstance(context.project).findFile(file) ?: continue
+            psiFile.apply {
+                if (jsonVersion) {
+                    this.findKeyForTextAsJson(text)?.let { return it }
+                } else {
+                    // todo: this.findKeyForTextAsLang(text)
+                }
+            }
```
LlamaLad7: Not sure I'd go for the `apply` over just `psiFile.findKeyForTextAsJson(text)`

LlamaLad7 (changes requested):

src/main/kotlin/translations/TranslationFiles.kt
Moulberry: [a135ee5](https://github.com/minecraft-dev/MinecraftDev/pull/2292/commits/a135ee54cdb3eb9d910636ea014f32906314292d)

src/main/kotlin/translations/intentions/ConvertToTranslationIntention.kt
Moulberry: [1ab56e9](https://github.com/minecraft-dev/MinecraftDev/pull/2292/commits/1ab56e9784fd3512976b4f0963019ada86ac9f3a)

src/main/kotlin/translations/TranslationFiles.kt
Moulberry: [271746a](https://github.com/minecraft-dev/MinecraftDev/pull/2292/commits/271746a3b0fea06c09f033fef527351451d25505)

src/main/kotlin/MinecraftSettings.kt
Moulberry: [97988d6](https://github.com/minecraft-dev/MinecraftDev/pull/2292/commits/97988d6171f106b430b48a6ee7eebedb0931add5)

src/main/kotlin/MinecraftConfigurable.kt
```diff
@@ -91,6 +92,16 @@ class MinecraftConfigurable : Configurable {
             }
         }
 
+        group(MCDevBundle("minecraft.setting.translation")) {
```
RedNesto: Nitpick, same for the two other keys below

```suggestion
        group(MCDevBundle("minecraft.settings.translation")) {
```

src/main/kotlin/MinecraftSettings.kt
```diff
@@ -37,6 +37,9 @@ class MinecraftSettings : PersistentStateComponent<MinecraftSettings.State> {
         var underlineType: UnderlineType = UnderlineType.DOTTED,
 
         var isShadowAnnotationsSameLine: Boolean = true,
+
+        var isForceJsonTranslationFile: Boolean = false,
+        var convertToTranslationTemplate: String = "net.minecraft.client.resources.I18n.format(\"\$key\")",
```
Earthcomputer: I believe intentions can have their own settings, so no need to make this a global setting. Although in this case we could probably get away with it being hardcoded if you use `MappingsManager`. Right now you would need to add it to `HardcodedYarnToMojmap` as we haven't fully implemented the mappings API yet.

It would be nice to figure out why we're unable to detect the MC version rather than forcing it. But this is probably good enough for a temporary solution.

Earthcomputer (changes requested):

Earthcomputer:
We have already dropped support for new features for MC 1.12, so I don't mind support for lang files being unimplemented.

src/main/kotlin/MinecraftSettings.kt
Moulberry: Its not just about mapping from yarn to mojmap, I personally like to use a small wrapper that can do some additional formatting. People on other platforms (paper, etc.) might also want to leverage the intention without being forced to use NMS. If you still want it to go through HardcodedYarnToMojmap let me know and I'll make the change

I'll look into how to move the settings into the intention settings

Earthcomputer: That's a fair enough reason, although I think the usage of yarn is common enough that we need to be able to make it work without any config changes. I guess just leave the remapping problem for now and we'll fix it once we get the mappings stuff up and running.

Moulberry: I'll try see if I can add an explicit checkbox for "Use Custom Template .. etc", which should make it easier to implement that behaviour in the future.

Moulberry: It was found that intentions cannot have their own settings. Leaving the options in MinecraftConfigurable seems like the easiest solution right now

src/main/kotlin/MinecraftConfigurable.kt
```diff
@@ -97,8 +99,17 @@ class MinecraftConfigurable : Configurable {
                 checkBox(MCDevBundle("minecraft.settings.translation.force_json_translation_file"))
                     .bindSelected(settings::isForceJsonTranslationFile)
             }
-            row(MCDevBundle("minecraft.settings.translation.convert_to_translation_template")) {
+
+            lateinit var property: ComponentPredicate
```
RedNesto: This variable deserves a better name.

Also not a fan of having this option in an application-wide configurable, its value will vary based on the project. It would be better to have it in a project-level configurable.

src/main/kotlin/MinecraftConfigurable.kt
Moulberry: [b3bd4e7](https://github.com/minecraft-dev/MinecraftDev/pull/2292/commits/b3bd4e76f7057f9dab92acc49449ad39e3f450a4)

src/main/kotlin/MinecraftConfigurable.kt
Moulberry: [d1050e8](https://github.com/minecraft-dev/MinecraftDev/pull/2292/commits/d1050e806c759f3bec305267b7e9b1bc58a2f2bf)

src/main/kotlin/translations/sorting/TranslationTemplateConfigurable.kt
```diff
@@ -61,6 +75,33 @@ class TranslationTemplateConfigurable(private val project: Project) : Configurab
...
+
+            row {
+                textField().bindText(translationSettings::convertToTranslationTemplate)
+                    .enabledIf(allowConvertToTranslationTemplate)
+            }
+
+            onApply {
+                for (project in ProjectManager.getInstance().openProjects) {
+                    ProjectView.getInstance(project).refresh()
+                }
```
RedNesto: Why is this needed?

src/main/kotlin/translations/TranslationFiles.kt
```diff
@@ -110,12 +111,46 @@ object TranslationFiles {
...
+            ?: throw IllegalArgumentException("Cannot add translation for element outside of module")
+        var jsonVersion = true
+        if (!TranslationSettings.getInstance(context.project).isForceJsonTranslationFile) {
+            val version =
+                context.mcVersion ?: throw IllegalArgumentException("Cannot determine MC version for element $context")
+            jsonVersion = version > MC_1_12_2
+        }
+
+        if (!jsonVersion) {
+            return null // todo: lang file support
```
RedNesto: Honestly I think you can ignore lang files, the support for that format is basically a dangling remnant of now unsupported versions.

Moulberry: [a6112d6](https://github.com/minecraft-dev/MinecraftDev/pull/2292/commits/a6112d6d6299f333762e64609930b7c71463b2c3)

src/main/kotlin/translations/sorting/TranslationTemplateConfigurable.kt
Moulberry: [9c4a0c0](https://github.com/minecraft-dev/MinecraftDev/pull/2292/commits/9c4a0c09a8260f79ed2a4c74ed5ff73ad0fc9d24)

src/main/kotlin/translations/intentions/ConvertToTranslationIntention.kt
```diff
@@ -73,12 +77,19 @@ class ConvertToTranslationIntention : PsiElementBaseIntentionAction() {
...
                 if (replaceLiteral) {
+                    val translationSettings = TranslationSettings.getInstance(project)
                     val psi = PsiDocumentManager.getInstance(project).getPsiFile(editor.document) ?: return
                     psi.runWriteAction {
                         val expression = JavaPsiFacade.getElementFactory(project).createExpressionFromText(
-                            "net.minecraft.client.resources.I18n.format(\"$key\")",
+                            if (translationSettings.isUseCustomConvertToTranslationTemplate) {
+                                translationSettings.convertToTranslationTemplate.replace("\$key", key)
+                            } else {
+                                "net.minecraft.client.resources.I18n.format(\"$key\")"
```
Earthcomputer, line 91: Wasn't the point of adding a checkbox for a custom template to use the mappings manager?

Earthcomputer (changes requested):
Looks good other than this one thing, and also formatting needs to be fixed with ktlint

src/main/kotlin/translations/intentions/ConvertToTranslationIntention.kt
Moulberry: My understanding from your previous message is that the mappings API isn't ready yet.
I added the checkbox so that when it's ready it can be integrated without requiring users to tick another box when they update.

RedNesto:
Thank you for your PR.

I rebased and merged it in 336924bed50a7b31ab01929da3477c5742cf9b02 with a change addressing Earth's comment.

PR #2293, opened 2024-05-07
Allow dynamic selectors to choose the class to perform member lookups in

Bawnorton:
## Motivation

MixinSquared's [`@TargetHandler`](https://github.com/Bawnorton/MixinSquared/wiki#target-handler) is a dynamic selector which targets merged handlers, the accepted references are the name of the mixin and the original name of the handler. The method targeting is the same as `@At.method` but the members are declared in the mixin class specified by the `mixin` attribute, thus, when looking up members for a `@TargetHandler` selector, the class specified in `mixin` would be the one to lookup. Mcdev's member lookups work off the target class specified in the `@Mixin` annotation, thus, this change is needed to allow the lookup to occur in a different class.

This pr also fixes namespace lookup as you do not have to declare the namespace for a dynamic selector in the annotation and can instead declare it at registration.

See #2287

### DIFF
--- a/src/main/kotlin/platform/mixin/reference/MixinSelectors.kt
+++ b/src/main/kotlin/platform/mixin/reference/MixinSelectors.kt
@@ -52,6 +52,7 @@ import com.intellij.openapi.util.text.StringUtil
 import com.intellij.psi.CommonClassNames
 import com.intellij.psi.JavaPsiFacade
 import com.intellij.psi.PsiAnnotation
+import com.intellij.psi.PsiCallExpression
 import com.intellij.psi.PsiClass
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiField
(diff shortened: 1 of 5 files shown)


src/main/kotlin/platform/mixin/reference/MixinSelectors.kt
```diff
@@ -387,6 +389,25 @@ private class MixinRegexSelector(
 
 // Dynamic selectors
 
+/**
+ * A dynamic selector can change which class to query for members if desired.
+ */
+interface DynamicMixinSelector : MixinSelector {
+    fun redirectOwner(owner: ClassNode): ClassNode {
```
Earthcomputer: I think it's fine for this method to be moved directly into the `MixinSelector` interface. That would remove the need for the static method below. Also maybe the method could be called `getCustomOwner` or something?

Earthcomputer (commented):
Looks good other than one thing.

Earthcomputer (approved):

Issue #2264, opened 2024-03-17
Outdated maven shade and compiler plugin versions


LB45440078L:
### Minecraft Development for IntelliJ plugin version

2023.3-1.7.3

### Description of the feature request

Maven shade and compile plugin are not up to date for bukkit, spigot, paper or bungeecord plugins.

Please update maven shade plugin to 3.5.2 and maven compile plugin to 3.12.1

Issue #2254, opened 2024-03-04
NeoForge template .gitignore fild error


TT432:
### Minecraft Development for IntelliJ plugin version

2023.3-1.7.3-470

### IntelliJ version

2023.3

### Operating System

Windows 11

### Description of the bug

NeoForge template's `.gitignore` file is old forge template's `.gitignore` file, NeoForge moved `runs` dir, and MDK update new `.gitignore` file

Issue #2252, opened 2024-03-02
Forge template always targeting latest Java version instead of the one specified


Prunoideae:
### Minecraft Development for IntelliJ plugin version

2023.1-1.6.12

### IntelliJ version

IntelliJ IDEA 2023.1.3 (Community Edition)

### Operating System

Windows 11

### Target platform

Minecraft Forge

### Description of the bug

When setting up a project from the template:

![image](https://github.com/minecraft-dev/MinecraftDev/assets/24620047/440274cb-f099-4349-bbc5-09e0bd51571a)

The template targets 18 instead of 17:

![image](https://github.com/minecraft-dev/MinecraftDev/assets/24620047/04bc7ab0-0ab3-46d6-b609-898da4b0e8d9)

While the Architectury template does not:

![image](https://github.com/minecraft-dev/MinecraftDev/assets/24620047/29fe2cc6-56fc-4294-9ee2-90a91300dae0)

![image](https://github.com/minecraft-dev/MinecraftDev/assets/24620047/1068ded4-eaf9-44da-a449-c26bb504442d)

Issue #2141, opened 2023-10-05
MinecraftDev's "New Project" Interface is EXTREMELY Laggy


TechnicallyCoded:
### Minecraft Development for IntelliJ plugin version

2023.2-1.6.10

### IntelliJ version

2023.2.2

### Operating System

MacOS 13.3.1

### Target platform

Bukkit, Spigot, Paper, Sponge, Architectury, Minecraft Forge, Fabric, Velocity, Mixins, BungeeCord, Waterfall

### Description of the bug

The interface for creating a new project is extremely slow and lags out the whole UI. Changing the type of project creation, makes it return to working perfectly normal. The issue is very likely directly caused by the MinecraftDev addon.

Screen recording of this happening below as I switch back and forth between Minecraft (1 fps) and whatever else (which doesn't lag at all):

https://github.com/minecraft-dev/MinecraftDev/assets/29813490/951d6763-58b5-480a-b81e-c3bfe5daa992

RedNesto:
This is very likely due to the fact we have a lot of fields that are pre-filled and updated based on the project name, where the other example you provided doesn't have any.

Also I just tested and it isn't nearly as bad on my end.

TechnicallyCoded:
Updating 4-5 text fields is unlikely to require 500ms-1000ms of processing no matter the circumstance.

DenWav:
I have experienced this a number of times myself. We really aren't doing anything crazy, all is the APIs we're using to build the UI are JetBrains, but we should look to find out if there's something we're doing that's slowing it down.

PR #2304, opened 2024-06-03
Repo-based creator templates

### LINKED ISSUE #2294: Add Split Source in Fabric

2024.1-1.7.5

Add config to select split source in Fabric

RedNesto:
### Why?

The main motivation is to offload the templates to a separate repo, which are updated independently when opening the creator for the first time in the IDE session. This means templates can be easily updated and tested in real time, no IDE restart (thank goodness.)

Also the creator file template settings became really cluttered and the UI is quite clunky sometimes. Now the templates are stored in a flat dir (zip archives are also supported.)

This also opens the door for anyone to create their own templates, and makes maintaining our templates so much easier (well, I might be biased on that.)

Oh yeah and it also gets rid of some "magic" code gen that breaks embarrassingly often for obscure reasons, so the fabric creator won't have this nasty issue where the main class is incomplete.

### Overview

The templates are still Velocity templates, but they are now put together by a JSON file describing a specific template. See the templates repo at https://github.com/minecraft-dev/templates.

Template descriptors mostly contain properties (most of which are displayed in the creator UI) and the template files.

All file templates have access to all the properties in the descriptor. Some string in the descriptor also have access to properties (file `template`, `destination` and `condition`.)

Properties cannot reference properties declared after them (they are setup one by one in a single step.) I chose this limitation so that we cannot easily cause recursion.

Some properties will be of a custom type, or expose an existing class of the plugin, all of those should be marked with `@TemplateApi` and we should pay attention to not break the source compatibility of those classes without a good reason.

~Used templates are remembered (only the path, not the files contents), there's no way to register them permanently (this could be added to the main settings eventually)~ Template repositories can be configured in the main configurable, recent templates are no longer remembered.

Fixes #2294 #2264 #2254 #2252 #2141 #2118 #1835 #804 #771


Earthcomputer:
I've only spent about 5 minutes looking at it so far, but one immediate question I have, is how are you going to make the options and labels translatable?

RedNesto:
Yeah I've been thinking about it, at least for strings in our resource bundle. Unsure what would be the best way to handle translations provided by the repository itself.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "templates"]
+	path = templates
+	branch = main
+	url = https://github.com/RedNesto/mcdev-templates
```
Earthcomputer: This'll need updating.

src/main/kotlin/creator/custom/providers/TemplateProvider.kt
```diff
@@ -0,0 +1,136 @@
...
+import javax.swing.JComponent
+
+/**
+ * Extensions responsible for creating a [TemplateDescriptor] based on whatever data it is provided in its [UI][setupUi].
+ */
+interface TemplateProvider {
+
+    fun getLabel(): String
+
+    fun getTooltip(): String? = null
```
Earthcomputer: Maybe these could be `val`s instead?

src/main/kotlin/creator/custom/CustomPlatformStep.kt
```diff
@@ -0,0 +1,357 @@
...
+
+        templateProviderPlaceholder.component = templateProvider.setupUi(context, propertyGraph, provideTemplate)
+    }
+
+    private fun initTemplates(
+        taskParentComponent: JComponent?
+    ) {
+        selectedTemplate = EmptyLoadedTemplate
+
+        val task = object : Task.Modal(
```
Earthcomputer: Is it a good idea from a UX perspective to use modal tasks inside an `AbstractLongRunningAssetsStep`, which already have a loading icon? I'm not sure. Also, are these modal tasks cancellable?

src/main/kotlin/creator/custom/CustomPlatformStep.kt
```diff
@@ -0,0 +1,357 @@
...
+                    continue
+                }
+
+                val fileName = destPath.fileName.toString().removeSuffix(".ft")
+                val baseFileName = FileUtilRt.getNameWithoutExtension(fileName)
+                val extension = FileUtilRt.getExtension(fileName)
+                val fileTemplate = CustomFileTemplate(baseFileName, extension)
+                fileTemplate.text = templateContents
+                assets.addAssets(GeneratorTemplateFile(projectPath.relativize(destPath).toString(), fileTemplate))
+            } catch (e: Exception) {
```
Earthcomputer: `Throwable` maybe?

src/main/kotlin/creator/custom/EvaluateTemplateExpressionAction.kt
```diff
@@ -0,0 +1,81 @@
...
+                field = textField().align(Align.FILL).component
+                field.isEditable = false
+            }
+        }
+
+        override fun doOKAction() {
+            val props = mapOf(
+                "BUILD_SYSTEM" to "gradle",
+                "USE_PAPER_MANIFEST" to false,
+                "MAIN_CLASS" to ClassFqn("io.github.rednesto.test.Test")
```
Earthcomputer, line 76: This probably isn't a good default for everyone

src/main/kotlin/creator/custom/finalizers/RunGradleTasksFinalizer.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.intellij.openapi.diagnostic.thisLogger
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.project.guessProjectDir
+
+class RunGradleTasksFinalizer : CreatorFinalizer {
+
+    override fun execute(project: Project, properties: Map<String, Any>, templateProperties: Map<String, Any?>) {
+        @Suppress("UNCHECKED_CAST")
+        val tasks = properties["tasks"] as? List<String>
+            ?: return
```
Earthcomputer: Do these properties come unvalidated from the json? If so, I think we are going to have to put some thought into what happens if the json is invalid:
- Log something if this cast fails.
- Check that every element is a string before doing the unchecked cast, otherwise it may throw hard-to-trace exceptions in the future.

Probably this needs to be looked for elsewhere too.

src/main/kotlin/creator/custom/model/CreatorJdk.kt
```diff
@@ -0,0 +1,31 @@
...
+package com.demonwav.mcdev.creator.custom.model
+
+import com.intellij.openapi.projectRoots.JavaSdk
+import com.intellij.openapi.projectRoots.Sdk
+
+@TemplateApi
+data class CreatorJdk(val sdk: Sdk?) {
+
+    val javaVersion: Int
+        get() = sdk?.let { JavaSdk.getInstance().getVersion(it)?.ordinal } ?: 17
```
Earthcomputer, line 30: Might want to put this 17 in a global default Java version constant somewhere, that we can bump as necessary. MC 1.20.6 is already on Java 21.

src/main/kotlin/creator/custom/model/TemplateApi.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.demonwav.mcdev.creator.custom.model
+
+/**
+ * Marker annotation indicating classes exposed to templates.
+ *
+ * Be careful of not breaking source or binary compatibility of those APIs without a good reason.
+ */
+annotation class TemplateApi
```
Earthcomputer, line 30: Is this enforced in any way or just documentation? Also, it should probably be `@Target(CLASS)` and with an appropriate `@Retention`.

src/main/kotlin/creator/custom/providers/BuiltInTemplateProvider.kt
```diff
@@ -0,0 +1,98 @@
...
+
+            val data = result.onError {
+                thisLogger().warn("Could not fetch builtin templates update", it)
+            }.getOrNull() ?: return
+
+            try {
+                val zipPath = PluginUtil.plugin.pluginPath.resolve("lib/resources/builtin-templates.zip")
+                zipPath.writeBytes(data)
+                FileUtil.deleteRecursively(builtinTemplatesPath)
+                ZipUtil.extract(zipPath, builtinTemplatesPath, null)
```
Earthcomputer: Hmm is it a good idea to overwrite installed plugin? I don't know, never thought about it before.

src/main/kotlin/creator/custom/types/IntegerCreatorProperty.kt
```diff
@@ -0,0 +1,92 @@
...
+    graph: PropertyGraph,
+    descriptor: TemplatePropertyDescriptor,
+    properties: Map<String, CreatorProperty<*>>
+) : SimpleCreatorProperty<Int>(graph, descriptor, properties) {
+
+    override fun createDefaultValue(raw: Any?): Int = raw as? Int ?: 0
+
+    override fun serialize(value: Int): String = value.toString()
+
+    override fun deserialize(string: String): Int = string.toIntOrNull() ?: 0
```
Earthcomputer, line 46: Maybe these `deserialize` functions could return a nullable value, so that if the jsons are invalid a warning could be logged? (and then `createDefaultValue` could be used)

src/main/kotlin/creator/custom/types/LicenseProperty.kt
```diff
@@ -0,0 +1,74 @@
...
+    override val graphProperty: GraphProperty<LicenseData> =
+        graph.property(createDefaultValue(descriptor.default))
+
+    override fun createDefaultValue(raw: Any?): LicenseData =
+        deserialize(raw as? String ?: License.ALL_RIGHTS_RESERVED.id)
+
+    override fun serialize(value: LicenseData): String = value.id
+
+    override fun deserialize(string: String): LicenseData =
+        LicenseData(string, License.byId(string)?.toString() ?: string, ZonedDateTime.now().year.toString())
```
Earthcomputer: Maybe licenses could be made extensible, doesn't have to be in this PR though.

Earthcomputer (changes requested):
Haven't tested it yet, but here are some thoughts having read through it. I'm assuming that you'll get rid of the old creator code once this is done?

Earthcomputer:
I also noticed that in the template repository, Fabric mod ids are validated to a minimum length of 1, whereas on Fabric mod ids have to be at least of length 2.

src/main/kotlin/creator/custom/providers/TemplateProvider.kt
RedNesto: Yeah I was considering making them vals when I wrote that. Guess that'll make it cleaner.

src/main/kotlin/creator/custom/CustomPlatformStep.kt
RedNesto: They are not cancellable (yet)
Also inheriting `AbstractLongRunningAssetsStep` was a very old decision, maybe I can get rid of it. Unsure if I can actually use its functionality the way template providers work, given providers call that modal task themselves based on their own UI inputs.

src/main/kotlin/creator/custom/CustomPlatformStep.kt
RedNesto: Yeah, should also make sure not to log/report cancellation exceptions.

src/main/kotlin/creator/custom/model/CreatorJdk.kt
RedNesto: > Might want to put this 17 in a global default Java version constant somewhere

Fair

> that we can bump as necessary. MC 1.20.6 is already on Java 21.

Well, this value is supposed to be the default as a fallback. I expect template authors to always provide a default in the descriptor.

src/main/kotlin/creator/custom/model/TemplateApi.kt
RedNesto: This is only meant for documentation. Will add Target and Retention 👌

src/main/kotlin/creator/custom/providers/BuiltInTemplateProvider.kt
RedNesto: The idea is to keep the builtin templates up to date, another way would be to store that in a separate directory and use it if it exists, which would work too.

src/main/kotlin/creator/custom/types/IntegerCreatorProperty.kt
RedNesto: This is one of the places I couldn't really figure out properly as it was written very early.
Your solution sounds good

src/main/kotlin/creator/custom/types/LicenseProperty.kt
RedNesto: Yes, I simply moved the licenses to the templates repo and reused existing code.
The ids and corresponding files could be provided in a shared descriptor that other descriptors inherit from.

RedNesto:
I'll look into improving the template selection, maybe split plugins and mods, and move the providers to the settings so we can have repos saved there and displayed in the creator.

src/main/kotlin/creator/custom/EvaluateTemplateExpressionAction.kt
RedNesto: This whole action was just an early test, I'll remove it.

src/main/kotlin/creator/custom/finalizers/RunGradleTasksFinalizer.kt
RedNesto: Validation could happen during the template setup, and if anything wrong happens during the execution either 
- a notification for template authors in case of a problem that has not been detected during the initial validation (if it ever happens)
- a warn/error, whichever makes sense if this is an issue with the implementation

Validation wasn't my concern early on, and this PR isn't done yet, it's functional but not fully ready (maybe I should make it a draft 🤔)

terminalsin:
This is a huge undertaking and I'm so grateful for the people contributing to making this happen. I'm looking forward to this PR being merged :)

Earthcomputer:
Hi, I've just tried this out, but it's saying there's no templates available:
image
Do you have any idea why this might be?

RedNesto:
@Earthcomputer Oh that's not supposed to be like that by default. In the settings page there should be a table of template repositories with a Built In default repo. I'll look into it.

Earthcomputer:
Ah yes. It seems the template repositories are empty by default.

RedNesto:
Should be fixed now

Earthcomputer:
Just a couple of things:
- Use Mixins should be enabled by default on the Fabric and Architectury platforms, and disabled by default on Forge and NeoForge. I've noticed that the value of this checkbox is connected on all platforms, maybe it should be disconnected (stored with a different key).
- `FabricApi` and `ArchitecturyApi` should not be in pascal case.
- While fetching versions from the internet, there could be an `AsyncProgressIcon`. In the old creator this is done inside `AbstractLatentStep`.

RedNesto:
Fixed all of these.

Also I forgot to answer to your question about the old wizard. 
I plan on keeping it until the end of the year, so people who have issues with this new one have a "backup". Guess I should add a warning that it will be removed at some point in the future in the UI.

src/main/kotlin/creator/custom/types/IntegerCreatorProperty.kt
RedNesto: So I think I'll keep it like that because `createDefaultValue` can take anything, usually it's a String, but it can be an Int, Boolean or even a Map if the property supports it.

The serialization methods are meant only for remembering values and have to be from/to Strings.

Deserialize can always fallback to the default value, I leave it up to the implementation to decide.

Earthcomputer (approved):

PR #2324, opened 2024-07-04
Fix/llamas expressions fixes

LlamaLad7:
Ticks off all the cards I had
See individual commits

### DIFF
--- a/src/main/kotlin/platform/mixin/expression/MEExpressionCompletionUtil.kt
+++ b/src/main/kotlin/platform/mixin/expression/MEExpressionCompletionUtil.kt
@@ -1178,6 +1178,11 @@ object MEExpressionCompletionUtil {
 
     private fun addDefinition(context: InsertionContext, id: String, definitionValue: String): PsiAnnotation? {
         val contextElement = context.file.findElementAt(context.startOffset) ?: return null
+        return addDefinition(contextElement, id, definitionValue)
+    }
+
+    fun addDefinition(contextElement: PsiElement, id: String, definitionValue: String): PsiAnnotation? {
+        val project = contextElement.project
         val injectionHost = contextElement.findMultiInjectionHost() ?: return null
         val expressionAnnotation = injectionHost.parentOfType<PsiAnnotation>() ?: return null
         if (!expressionAnnotation.hasQualifiedName(MixinConstants.MixinExtras.EXPRESSION)) {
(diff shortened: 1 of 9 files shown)


src/main/kotlin/platform/mixin/expression/MEExpressionCompletionUtil.kt
```diff
@@ -1271,13 +1275,25 @@ object MEExpressionCompletionUtil {
...
+                val callExpr = factory.createExpression("super.?()") as MESuperCallExpression
+                expression.memberName?.let { callExpr.memberName!!.replace(it) }
+                variants += callExpr
+            }
         }
 
         return variants
     }
 
+    private fun createUniqueLookup(text: String) = LookupElementBuilder.create(Any(), text)
```
Earthcomputer, line 1294: I'm a bit concerned that this looks a bit like a hack. Is the first parameter used for anything important?

src/main/kotlin/platform/mixin/expression/MEExpressionCompletionUtil.kt
```diff
@@ -1179,6 +1178,11 @@ object MEExpressionCompletionUtil {
 
     private fun addDefinition(context: InsertionContext, id: String, definitionValue: String): PsiAnnotation? {
         val contextElement = context.file.findElementAt(context.startOffset) ?: return null
+        return addDefinition(contextElement, id, definitionValue)
+    }
+
+    fun addDefinition(contextElement: PsiElement, id: String, definitionValue: String): PsiAnnotation? {
+        val project = contextElement.project
```
Earthcomputer: We should keep the project as a parameter if possible, as IIRC accessing the project from a PSI element is less performant. Although it probably doesn't matter here.

src/main/kotlin/platform/mixin/expression/MEExpressionCompletionUtil.kt
LlamaLad7: It's used by custom sorters but we don't have any of those. Otherwise it's just used to determine the uniqueness.

src/main/kotlin/platform/mixin/expression/MEExpressionCompletionUtil.kt
LlamaLad7: It got the project from the element prior to my changes

LlamaLad7: Oh sorry it was from the `InsertionContext`, I see.

Earthcomputer (approved):

PR #2340, opened 2024-07-30
Use `LayeredIcon` for mixin classes

kikugie:
## Motivation
Newly introduced mixin files icons look awkward - pixelation becomes noticable, they don't fit with the Intellij icon style and don't convey the information about the type of the class:  
image

## Solution
Intellij supports layered icons, similar to Kotlin classes, which are more visually pleasing:

### New UI - dark
image

### Old UI - dark
image


src/main/kotlin/platform/mixin/framework/MixinIconProvider.kt
```diff
@@ -21,15 +21,21 @@
...
-        PlatformAssets.MIXIN_ICON.takeIf {
-            MinecraftSettings.instance.mixinClassIcon && element is PsiClass && element.isMixin
-        }
+    override fun getIcon(element: PsiElement, flags: Int) : Icon? =
+        if (MinecraftSettings.instance.mixinClassIcon && element is PsiClass && element.isMixin)
+            LayeredIcon.create(element.withoutMixin().getIcon(flags), MixinAssets.MIXIN_MARK)
+        else null
+
+    private fun PsiClass.withoutMixin(): PsiClass =
+       (copy() as PsiClass).apply { mixinAnnotation?.delete() }
```
RedNesto: This is rather hacky, I think using `IconLayerProvider` is more appropriate than `IconProvider` in that case.

kikugie: I wasn't aware of that - pushed an update

src/main/resources/assets/icons/mixin/mixin_mark.svg
```diff
@@ -0,0 +1,215 @@
...
+       stroke="none"
+       fill="#ebac24"
+       fill-rule="evenodd"
```
RedNesto: Looks like only this path is needed to render correctly, there's like 200 lines of fat for no clear reason.

kikugie: Weird that Inkscape was keeping all the hidden layers I used for testing

kikugie:
Excuse the impatience, does this look good? (The old icon is irritating a bit)

RedNesto (approved):

Earthcomputer (approved):

PR #2350, opened 2024-08-01
Create obfuscation-explorer project

DenWav:
This is a separate plugin to support all different Java mapping formats. This plugin will provide a extension points for the Minecraft Development plugin to hook into and provide mapping information. It will also provide a generic settings page to allow users to configure their own mappings for the project.

As is this is incomplete, not ready to be used for anything really. However, it's a big change that doesn't currently change the base Minecraft Development plugin at all, so it's safe to merge to make it easier to work on going forward.


RedNesto (approved):
This looks good to me for what's there.

Would be nice to separate those gradle and mcdev changes into their own commits out of this PR though. I'd like to backport the intellij gradle plugin 2.0 upgrade to dev and this would cause many merge conflicts in this PR.

PR #2356, opened 2024-08-10
Add inspection for targeting the same class multiple times in a mixin

O7410:
Currently it's a weak warning

### DIFF
--- a/src/test/kotlin/platform/mixin/MixinDuplicateTargetInspectionTest.kt
+++ b/src/test/kotlin/platform/mixin/MixinDuplicateTargetInspectionTest.kt
@@ -1,0 +1,350 @@
+/*
+ * Minecraft Development for IntelliJ
+ *
+ * https://mcdev.io/
+ *
+ * Copyright (C) 2024 minecraft-dev
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, version 3.0 only.
+ *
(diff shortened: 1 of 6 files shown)


RedNesto:
This is looking pretty good, there's a few things to note:
- this does not cover cases where a class literal and string target both reference the same class
- it would be nice to have some tests for this inspection, you can use [`InvalidInjectorMethodSignatureInspectionTest`](https://github.com/minecraft-dev/MinecraftDev/blob/6b0471e93679cb6cd2b0d1d9c39409c1cbf811ca/src/test/kotlin/platform/mixin/InvalidInjectorMethodSignatureInspectionTest.kt) as a reference.
- you should be fine getting `PsiClass.mixinTargets` once and pass it around, instead of hitting the cache again and again

O7410:
Take a look at the latest commit

src/test/kotlin/platform/mixin/MixinDuplicateTargetInspectionTest.kt
```diff
@@ -0,0 +1,350 @@
...
+
+    private fun doTest(@Language("JAVA") code: String) {
+        buildProject {
+            dir("test") {
+                java("TestMixin.java", code)
+            }
+        }
+
+        fixture.enableInspections(MixinDuplicateTargetInspection::class)
+        fixture.checkHighlighting(false, false, true)
```
RedNesto, line 42: You can add one last `true` argument to remove the undesired unresolved highlights.

PR #2357, opened 2024-08-12
feat: add bedrock nbt and leveldat support

CoolLoong:
This PR adds support for `Little Endian` and `Network Varint` formatted NBT in Bedrock Edition. It also includes support for Bedrock Edition world `.dat` files. I'm not very familiar with `Kotlin`, so any suggestions for improvements are welcome.

### DIFF
--- a/src/main/kotlin/nbt/Nbt.kt
+++ b/src/main/kotlin/nbt/Nbt.kt
@@ -21,36 +21,99 @@
 package com.demonwav.mcdev.nbt
 
 import com.demonwav.mcdev.asset.MCDevBundle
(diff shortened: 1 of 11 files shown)


DenWav:
Thanks so much for you contribution! I sincerely apologize for the lack of response on this, life is hectic and I sometimes take breaks from my projects. This is something I've wanted to implement for a while so I'm excited to look at what you've put together.

Don't worry about conflicts or fixing tests, I'll take care of all that for you (unless you really want to).

CoolLoong:
Closed this PR because the current implementation is not very good. 
It's not very elegant in handling convert between different format.

PR #2358, opened 2024-08-14
Support split strings within inspections & method target references

FxMorin:
### Description:
Currently some inspections don't support strings that have been split to support a line width limit, same goes for Injection method targets.

### Changes:
Removed multiple direct `PsiLiteral` checks which where not needed, since all calls would lead to `PsiElement.constantStringValue`, which will return `null` if its not suppose to. 

Now, all these calls will use the `ConstantExpressionVisitor.compute` call which takes care of the split string logic.

#### Method Reference
Before - Single _(Expected behaviour)_
Before - Single
Before - Split
Before - Split

After - Split
After - Split

#### AmbiguousReferenceInspection
Before
Before

After
After

#### InvalidMemberReferenceInspection
Before
Before

After
After

#### UnnecessaryQualifiedMemberReferenceInspection
Before
Before

After
After

### DIFF
--- a/src/main/kotlin/platform/mixin/inspection/reference/AmbiguousReferenceInspection.kt
+++ b/src/main/kotlin/platform/mixin/inspection/reference/AmbiguousReferenceInspection.kt
@@ -35,7 +35,6 @@ import com.intellij.psi.PsiAnnotation
 import com.intellij.psi.PsiAnnotationMemberValue
 import com.intellij.psi.PsiArrayInitializerMemberValue
 import com.intellij.psi.PsiBinaryExpression
-import com.intellij.psi.PsiLiteral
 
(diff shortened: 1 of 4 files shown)


Earthcomputer (changes requested):
Are you sure that checks for `PsiLiteral` weren't the only problem? The `constantEvaluationHelper` is already supposed to handle string concatenation, see `ConstantExpressionVisitor.compute`

FxMorin:
> Are you sure that checks for `PsiLiteral` weren't the only problem? The `constantEvaluationHelper` is already supposed to handle string concatenation, see `ConstantExpressionVisitor.compute`

I was not aware of this. I just tested it, and you're right. I'll remove the changes to `PsiElement.constantStringValue`

Earthcomputer (approved):

PR #2365, opened 2024-08-29
AT Overhaul

### LINKED ISSUE #183: Add inspection for duplicate AT entries

### LINKED ISSUE #219: AccessTransformer usage not detected when using it for overriding a method
As said in the title

### LINKED ISSUE #2371: "No SRG name found" when using "Copy / Paste Special... > AT Entry"

2024.2-1.8.1

### IntelliJ version

2024.2.1 / 242.21829.142

### Operating System

Linux 6.10.9-arch1-1

### Target platform

_No response_

### Description of the bug

Target platform: NeoForge, it wasn't in the drop down.

When I use the "Copy / Paste Special... > AT Entry" context action on any MC code from NeoForge, nothing is copied and "No SRG name found" is shown. NeoForge 1.21 does not use SRG names at all in ATs, so this makes sense. The entry should be entirely based off of the Mojang mappings.

This can be reproduced easily using any NeoForge project, adding:
```
        ChunkHolder holder = null;
        holder.changedBlocksPerSection[0] = null;
```
and trying to copy an AT entry for `changedBlocksPerSection`.

RedNesto:
Fixes #2371, fixes #219, fixes #183


src/main/kotlin/platform/mcp/at/inspections/AtInspectionSuppressor.kt
```diff
@@ -0,0 +1,145 @@
...
+                return true
+            }
+        }
+
+        val file = element.containingFile as AtFile
+        return file.headComments.any { comment -> isSuppressing(comment.text, toolId) }
+    }
+
+    private fun isSuppressing(entryComment: String, toolId: String): Boolean {
+        val suppressed = entryComment.substringAfter("Suppress:").substringBefore(' ').split(',')
```
DenWav, line 53: To follow the IntelliJ standard shouldn't this look for `noinspecetion`?

src/main/kotlin/platform/mcp/at/inspections/AtInspectionSuppressor.kt
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    class AtSuppressQuickFix(element: PsiElement, val toolId: String) :
+        LocalQuickFixOnPsiElement(element), SuppressQuickFix {
+
+        override fun getText(): @IntentionName String = when (startElement) {
+            is AtEntry -> "Suppress $toolId for entry"
+            is AtFile -> "Suppress $toolId for file"
+            else -> "Suppress $toolId"
```
DenWav, line 81: Is there an option to suppress an individual entry inspection for the whole file? Or am I just not reading this right?

src/main/kotlin/platform/mcp/at/at-utils.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.demonwav.mcdev.util.MinecraftVersions
+import com.demonwav.mcdev.util.SemanticVersion
+
+fun MinecraftFacet.usesSrgMemberNames(): Boolean {
+    if (!this.isOfType(NeoForgeModuleType)) {
+        return true
+    }
+
+    val mcpModule = this.getModuleOfType(McpModuleType) ?: return true
+    val mcVersion = mcpModule.getSettings().minecraftVersion?.let(SemanticVersion::tryParse) ?: return true
```
DenWav: These default `return` values seem suspicious to me. I don't know if it matters really but this might make more sense as a `TriState`, with the above options returning `UNKNOWN`. Or I guess just `null` with `Boolean?`.

src/test/kotlin/framework/test-util.kt
```diff
@@ -131,3 +133,59 @@ fun testInspectionFix(fixture: JavaCodeInsightTestFixture, basePath: String, fix
...
+        val message = """|
+      |Expecting actual:
+      |  $actual
+      |to contain exactly in any order:
+      |  $expected
+      |elements not found:
+      |  $notFound
+      |and elements not expected:
+      |  $notExpected
+    """.trimMargin()
```
DenWav: Not sure if ktlint messed this up, looks like it needs to be indented by 1.

src/main/kotlin/platform/mcp/at/inspections/AtInspectionSuppressor.kt
RedNesto: I don't think there's a standard, the Properties plugin uses `suppress` as a keyword too.

src/main/kotlin/platform/mcp/at/inspections/AtInspectionSuppressor.kt
RedNesto: The fix supports both file-level and entry-level suppressions.

src/main/kotlin/platform/mcp/at/at-utils.kt
RedNesto: I agree, I used a simple Boolean to make it easier to use and have to make a decision anyway, and figured defaulting to using SRG is good enough. What would we do in case it is `UNKNOWN`?

DenWav: For this code that uses it, probably the same behavior. But it may be possible that later code will actually care, and doesn't really cost us anything to go ahead and do that.

Leclowndu93150:
Needs to be merged ngl

PR #2375, opened 2024-09-18
Add json5 support to mixin config jsons

Bawnorton:
Allow mixin config json inspections and references to recognise mixin configs of the json5 file type

## Motivation

[Stonecutter](https://github.com/kikugie/stonecutter), a gradle pre-processor plugin has been gaining traction for it's use in multi-loader, multi-version projects, and a plugin that was developed alongside it, [j52j](https://github.com/kikugie/j52j), which converts json5 files to json during compilation allowing versioned comments within json files resulting in json5 being used as the file type for a mixin config:

<https://github.com/Bawnorton/BetterTrims/blob/8558c398e0cd5a85fdd5dce6b2ed9e410d239743/src/main/resources/bettertrims.mixins.json5#L6-L12>

mcdev doesn't recognise json5 as a valid file type for mixin config jsons, so this pr adds that support.

### DIFF
--- a/src/main/kotlin/platform/mixin/config/reference/MixinConfigReferenceContributor.kt
+++ b/src/main/kotlin/platform/mixin/config/reference/MixinConfigReferenceContributor.kt
@@ -33,7 +33,6 @@ import com.intellij.psi.PsiReferenceRegistrar
 
 class MixinConfigReferenceContributor : PsiReferenceContributor() {
     override fun registerReferenceProviders(registrar: PsiReferenceRegistrar) {
-        println(this::class.qualifiedName)
         val jsonPattern = PlatformPatterns.psiElement(JsonStringLiteral::class.java)
             .inFile(PlatformPatterns.psiFile().withFileType(StandardPatterns.`object`(MixinConfigFileType.Json)))
         val json5Pattern = PlatformPatterns.psiElement(JsonStringLiteral::class.java)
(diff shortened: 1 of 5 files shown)


RedNesto (changes requested):
I don't think all of those changes are needed at all. Simply changing `filenameRegex` by adding `5?` at the end of the extension should do the job.

Bawnorton:
I tested that initially and it causes a json language inspection in the mixin json for violating the json standard as the config file type specifies that the language is Json not Json5

Screenshot 2024-09-19 at 3 24 17 PM

src/main/kotlin/platform/mixin/config/reference/MixinConfigReferenceContributor.kt
```diff
@@ -26,16 +26,23 @@ import com.demonwav.mcdev.util.isPropertyValue
...
 import com.intellij.patterns.StandardPatterns
 import com.intellij.psi.PsiReferenceContributor
 import com.intellij.psi.PsiReferenceRegistrar
 
 class MixinConfigReferenceContributor : PsiReferenceContributor() {
-
     override fun registerReferenceProviders(registrar: PsiReferenceRegistrar) {
-        val pattern = PlatformPatterns.psiElement(JsonStringLiteral::class.java)
-            .inFile(PlatformPatterns.psiFile().withFileType(StandardPatterns.`object`(MixinConfigFileType)))
+        val jsonPattern = PlatformPatterns.psiElement(JsonStringLiteral::class.java)
```
RedNesto: This isn't a great solution, a simpler way to do that is using an `or` pattern:

```kotlin
        val anyMixinConfigFileType = StandardPatterns.or(
            StandardPatterns.`object`(MixinConfigFileType.Json),
            StandardPatterns.`object`(MixinConfigFileType.Json5)
        )
        val pattern = PlatformPatterns.psiElement(JsonStringLiteral::class.java)
            .inFile(PlatformPatterns.psiFile().withFileType(anyMixinConfigFileType))
```

RedNesto (changes requested):

RedNesto (approved):
Looks good now, thanks for the PR 👍

PR #2383, opened 2024-10-03
Added local idea cersion for straighter updates on new EAP versions

CarinaSchoppe:
Whenever JetBrains creates an EAP version adding the plugin to the IDE results in an error because of the lack of versionsupport.
With the new idea version tag in the plugin#xml file its possible to easy update the existing plugin without any guarantee of version support just with the possibility to get it up running on the machines.


RedNesto:
The situation of EAP support is not that straightforward.

We use a lot of different and sometimes obscure APIs, some things break pretty much every time the first EAP of a new major version releases, and we usually release a working version for new EAPs the same day (or the next day depending on timezones) so this is really not needed.

PR #2390, opened 2024-11-03
Add moduleName in ClassFqn

Tuchan:
Useful for whenever you wanna create multiple modules inside of one project or just getting the lowerCamelCase name of the plugin

### DIFF
--- a/src/main/kotlin/creator/custom/model/ClassFqn.kt
+++ b/src/main/kotlin/creator/custom/model/ClassFqn.kt
@@ -43,6 +43,11 @@ data class ClassFqn(val fqn: String) {
      */
     val packagePath by lazy { packageName.replace('.', '/') }
 
+    /**
+     * The module name, which is the last part of the packageName
+     */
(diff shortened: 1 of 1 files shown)


RedNesto:
`PROJECT_NAME` is already the name of the module if you create a module in an existing project.

I'm not certain what one would use the lowerCamelCase name for, `MOD_ID`, `MOD_NAME` and `BUILD_COORDS.artifactId` are already there though.

if there really is a need for loweCamelCase this could be implemented as a string derivation.

Tuchan:
> I'm not certain what one would use the lowerCamelCase name for

lowerCamelCase can be used if for example, you wanna have a variable in code that takes the name of the plugin and adds `Api` at the end of it: `CrazyPlugin` -> `crazyPluginApi`
Nitpick... i know 🙄

> if there really is a need for loweCamelCase this could be implemented as a string derivation.

Will use string derivatives from now on.

Thanks 🙏

PR #2425, opened 2025-01-08
Move non-private static error to the access modifier (when there is one)

O7410:
before: entire line highlighted in red:
image
after: only the keyword (when there is one) is highlighted:
image


src/main/kotlin/platform/mixin/inspection/StaticMemberInspection.kt
```diff
@@ -33,6 +34,7 @@ import com.intellij.psi.PsiField
 import com.intellij.psi.PsiMember
 import com.intellij.psi.PsiMethod
 import com.intellij.psi.PsiModifier
+import org.jetbrains.kotlin.j2k.accessModifier
```
Earthcomputer: This will break when the Kotlin plugin is not installed. Please find an alternative API.

Earthcomputer (changes requested):
Not sure if `accessModifier` already does this, but in case it doesn't, you should make sure it always lands on the *access* modifier and not on other modifiers such as `static`.

Earthcomputer (approved):

PR #2442, opened 2025-02-19
Fix AsmDfaUtil analysis crash

LlamaLad7:
This fixes mixinextras expression analysis not working in some cases.

It would previously return `null` when asked for the supertype of an array type, causing a crash later on when `isAssignableFrom` received a null type. Instead, we match the reflection API used by ASM by default and return `Object` as the direct superclass of any array type.


Earthcomputer:
Thanks, can we also catch the exception in AsmDfaUtil, log an error (which will show up as an error in the bottom right of the screen) and return `null` in case there is something else?

Earthcomputer:
Never mind, looks like it was already doing that

PR #2457, opened 2025-04-09
Feature: region files support

### LINKED ISSUE #1798: support opening mca file.
I want to read mca file so can you support opening mca file as a set of nbt?
The mca file is made of multiple nbt.
I think it's one good to have a drop-down list of region and show nbt tag of selected region.

edgarogh:
Hello, I implemented a feature that closes #1798. I did this for my personal use but this PR is spontaneous, I can understand that this is not wanted as I've never got the chance to discuss that with a maintainer. No hard feelings if you aren't interested.

### Feature breakdown

- Ability to expand region files as if they were archives (which they somewhat are, to be fair)
  image
  - Only chunks that have been initialized (world-generated) are shown
  - Since region files should only contain NBT, and since the chunks are only indexed by their XY position, the file names are completely arbitrary. I chose to use the traditional `c.x.z.nbt` naming, but this can be changed.
  - These NBT files can obviously be opened in the already-existing NBT viewer (read-only though)
  - Due to the sheer simplicity of the region format, the parser was implemented directly in my code
- The chunk coordinates inside the files are _absolute_ (or relative to the entire world, if you prefer)
  image
  - This makes searching for a specific chunk much easier
  - The name of the MCA file is taken into consideration to know the location of the whole region itself
  - If the MCA file has an invalid name, I'm showing relative coordinates instead (I'm using tildes, as most Minecraft players would understand them as meaning "relative")
    image
- This is a bit unrelated, but this PR removes the "Save" button in the toolbar of NBT files if the file is read-only, which is important in this case -- as my code doesn't handle writing to region files -- but it might as well improve the compatibility with other extensions in scenarios where a read-only NBT file is opened
- NBT payloads inside region files are compressed using an algorithm specified within the region file (outside of the NBT payload, if that makes sense). If I didn't do anything special to handle that, the NBT editor's combo-box would always show "Uncompressed" because my code is handling the decompression before handing over the NBT payload to IntellIJ. I added a special case for that, so that the correct compression algorithm is shown. Because editing is not supported, the combobox is disabled anyway. To be frank, I'm not really happy with my code, it feels spaghetti-esque, but I'm willing to improve the structure if you have suggestions.
  image
  - Note: region files support algorithms that aren't typically used in typical standalone NBT files, namely ZLib and LZ4. I've made sure that they don't appear in the combobox of standalone NBT files

### What remains to be done

I prefer opening this PR before I spend too much time on the feature, as it is already sufficient for what I needed, but there are a few more things to do before it's mergeable.

- I did not bother with threads at all for the moment, i.e. the files are read on the main thread. The index of region files (the part needed to enumerate the chunk list) fits in the first 2KiB of the file, so should be really fast to read, but it would undeniably be better to do this properly, with threads. Reading the actual chunk content uses exclusively `InputStream`s that I'm handing back to IntelliJ, so I would assume that they're doing things correctly in terms of parallelism
- I'm not really handling parse errors. At best, I'm just returning eagerly from functions when something looks off, but my code probably deserves better error reporting.
- I haven't extensively tested my code.

### Disclosure

1. This is my first time ever using the IntelliJ SDK, I can't guarantee that everything I did is correct or idiomatic. I have decent (hobbyist) experience in Kotlin, though.
2. While I did not copy code from them, I wouldn't have been able to implement the custom archive type without the help of https://github.com/b3er/idea-archive-browser and their code being open source (Apache-2.0). They probably deserve a little bit of credit somewhere. Maybe in my own files, in a comment.

---

Please tell me if you're interested by this and if you have any remarks. I'll continue improving the code only if this is useful to others.
Have a great day!

### DIFF
--- a/src/main/kotlin/MinecraftTreeStructureProvider.kt
+++ b/src/main/kotlin/MinecraftTreeStructureProvider.kt
@@ -1,0 +1,47 @@
+/*
+ * Minecraft Development for IntelliJ
+ *
+ * https://mcdev.io/
+ *
+ * Copyright (C) 2025 minecraft-dev
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, version 3.0 only.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.demonwav.mcdev
+
+import com.demonwav.mcdev.region.RegionFileType
+import com.demonwav.mcdev.region.RegionPsiFileNode
+import com.intellij.ide.projectView.TreeStructureProvider
+import com.intellij.ide.projectView.ViewSettings
+import com.intellij.ide.projectView.impl.nodes.PsiFileNode
+import com.intellij.ide.util.treeView.AbstractTreeNode
+
+private fun mapMcaNode(node: AbstractTreeNode<*>): AbstractTreeNode<*> {
+    if (node is PsiFileNode) {
+        val value = node.value
+        if (value?.fileType is RegionFileType) {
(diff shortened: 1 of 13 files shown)


DenWav:
This is really cool. It's something I've wanted to for a long time. I have struggled to stay on top of this with this project due to other obligations, but I'm trying to make more of an effort now, so I apologize for such a long delay in responding. I'd like to see what we can do about getting this done and across the finish line.

edgarogh:
No worries, tell me if you'd like any changes, or just build on top of my branch if you want 😉

PR #2470, opened 2025-05-18
Fix/mixin fixes

LlamaLad7:
Various Mixin/MixinExtras fixes, see individual commits.

I had a look into the issue with `@Shadow` completion suggestions having their priority way too high but I didn't get that far, would appreciate if someone else could take a look at that one.


LlamaLad7:
Tests pass locally and I can't see any details in the logs that would indicate what's wrong...

Earthcomputer (approved):
Don't worry about the test failure, our tests have been failing intermittently lately

PR #2471, opened 2025-05-25
More Mixin Fixes

LlamaLad7:
Fixes mixin shadow autocomplete results often being prioritized over regular java ones
Also adds much better support for injection point specifiers.
The latter required some decently big reworks to the `CollectVisitor` system.
During that I greatly improved the (previously nonexistent due to bugs) message about which filters were responsible for the failure.
Happy to debate any of the refactors if you disagree with them.

Also as a side note, I keep my commit history clean, so it would be preferable for my PRs not to be squashed.

### DIFF
--- a/src/main/kotlin/platform/mixin/completion/InjectionPointTypedHandlerDelegate.kt
+++ b/src/main/kotlin/platform/mixin/completion/InjectionPointTypedHandlerDelegate.kt
@@ -1,0 +1,44 @@
+/*
+ * Minecraft Development for IntelliJ
+ *
+ * https://mcdev.io/
+ *
+ * Copyright (C) 2025 minecraft-dev
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, version 3.0 only.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.demonwav.mcdev.platform.mixin.completion
+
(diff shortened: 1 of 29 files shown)


src/main/kotlin/platform/mixin/completion/InjectionPointTypedHandlerDelegate.kt
```diff
@@ -0,0 +1,44 @@
...
+            return Result.CONTINUE
+        }
+        AutoPopupController.getInstance(project).autoPopupMemberLookup(editor) {
+            val offset = editor.caretModel.offset
+            val element = it.findElementAt(offset)?.findContextElement()
+            InjectionPointReference.ELEMENT_PATTERN.accepts(element)
+        }
+        return Result.CONTINUE
+    }
+}
```
Earthcomputer: Newline at end of file

src/main/kotlin/platform/mixin/util/InjectionPointSpecifier.kt
```diff
@@ -0,0 +1,48 @@
...
+            val isInsideSlice = at.parentOfType<PsiAnnotation>()?.hasQualifiedName(SLICE) == true
+            return isInsideSlice
+        }
+
+        private fun isAllowedOutsideSlice(project: Project): Boolean =
+            (JavaPsiFacade.getInstance(project)
+                .findClass(SPECIFIER, GlobalSearchScope.allScope(project))
+                != null)
+    }
+}
```
Earthcomputer: Newline at end of file

Earthcomputer:
The `InjectionPoint` code is API, so we need to make sure our API consumers, notably MixinSquared, are aware of these changes

LlamaLad7:
As in I should inform them / make a PR there, or try to maintain bin compat?

Earthcomputer:
Trying to maintain binary compat is the best, but not at the expense of keeping our own codebase maintainable. We don't have many API consumers so we can get away with breaking binary compat as long as we inform the major ones (I can only think of MixinSquared tbh)

LlamaLad7:
I can inform them yeah, they don't have any injection points anyway so should be minor

LlamaLad7:
The MixinSquared plugin needs no changes.

Earthcomputer (approved):

PR #2494, opened 2025-07-18
Feature/flow graph improvements

LlamaLad7:
I would squash when merging.

Greatly improves the code style and functionality of MixinExtras Expression Flow Graphs.
image
There are some more improvements, see individual commits, but the big news is that you can now right click an expression to debug it and see exactly why it isn't matching.


LlamaLad7:
Tests pass locally.

Earthcomputer (approved):
¯\\\_(ツ)\_/¯

PR #2495, opened 2025-07-19
Fix: Trigger completion properly in MEExpressionCompletionTest.

LlamaLad7:
Hopefully this will fix the race condition.


LlamaLad7:
Promising

Earthcomputer:
May or may not work as it's intermittent but we can try it and see

Earthcomputer:
Nope didn't work lol

LlamaLad7:
Shame, best to avoid impl classes anyway.

PR #2538, opened 2025-11-08
Update AT Action to work with NeoForge 1.21.4+ and Most likely Mojangs Non-Obfuscated Source post 1.21.11

Nyfaria:
make work with neoforges runtime mappings (will most likely work in the future for mojangs non-obfuscated source post 1.21.11)


Earthcomputer:
I'm sorry but this was not the right way to fix it either. `SrgActionBase` is an abstraction over multiple SRG actions, you have assumed in the base class a particular implementation but this is not the case. I am closing this PR in favor of a correct implementation I have just made: https://github.com/minecraft-dev/MinecraftDev/commit/df97f78630a5f3da8711818534bd6d3e86f3557b

PR #2559, opened 2025-12-19
Fix: Resolve LVTs sometimes being declared present when they are not.

LlamaLad7:
This was noticeable when autocompleting Expressions.


Earthcomputer:
I've got no idea why this works but I assume it does

Earthcomputer (approved):

Issue #2369, opened 2024-09-02
Local Classes should error the same as Inner Classes


ChiefArug:
### Minecraft Development for IntelliJ plugin version

2024.1-1.8.1

### Description of the feature request

There is currently an error message when trying to use an inner class in a mixin that is not also a `@Mixin` class itself.
![image](https://github.com/user-attachments/assets/88c8923d-5b70-420e-bd8b-9df9fb1b8ed0)
This does not show on method local classes though, even though the same restrictions apply (attempting to use such a class results in `org.spongepowered.asm.mixin.transformer.throwables.IllegalClassLoadError: <localclass> is in a defined mixin package <mixin.package> owned by <mixin.config> and cannot be referenced directly
`)
![image](https://github.com/user-attachments/assets/5d0d572b-7ba6-4ac6-826b-8a8dff55b1d4)

PR #2561, opened 2025-12-22
Fix: Allow Local classes and inner classes in mixins.

LlamaLad7:
They in fact are allowed.


Earthcomputer:
@LlamaLad7 this code was added to address https://github.com/minecraft-dev/MinecraftDev/issues/2369, has the situation changed since?

LlamaLad7:
I don't think it was ever the case. Mixin does not attempt to make any distinction between anonymous and named inner classes, they're handled identically.

There *is* a bug where Mixin will not remap field accesses to inner classes (which will soon be fixed in FabricMixin) but that affects anonymous classes just the same.

PR #2574, opened 2026-01-09
feat: prepare for plugin template update

Strokkur424:
This PR adds required changes to the main plugin in preparation for https://github.com/minecraft-dev/templates/pull/42/.

### DIFF
--- a/src/main/kotlin/creator/custom/derivation/ExtractPaperApiVersionPropertyDerivation.kt
+++ b/src/main/kotlin/creator/custom/derivation/ExtractPaperApiVersionPropertyDerivation.kt
@@ -23,37 +23,18 @@ package com.demonwav.mcdev.creator.custom.derivation
 import com.demonwav.mcdev.creator.custom.PropertyDerivation
 import com.demonwav.mcdev.creator.custom.TemplateValidationReporter
 import com.demonwav.mcdev.creator.custom.types.CreatorProperty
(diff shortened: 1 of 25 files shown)


src/main/kotlin/creator/custom/derivation/ExtractPaperApiVersionPropertyDerivation.kt
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        if (from.parts.size == 3) {
+          val (part1, part2, part3) = from.parts
+          if (part1 is SemanticVersion.Companion.VersionPart.ReleasePart &&
+            part2 is SemanticVersion.Companion.VersionPart.ReleasePart &&
+            part3 is SemanticVersion.Companion.VersionPart.ReleasePart
+          ) {
+            // From Minecraft version 1.20.5 onwards, the Paper API version also contains the 'minor' number.
+            if (part1.version >= 26 || (part2.version >= 21) || (part2.version == 20 && part3.version >= 5)) {
```
Earthcomputer: This is a very complicated version check, why not just compare with `MinecraftVersions.MC1_20_5`?

Strokkur424: Because I wasn't aware of that classes existence, thanks for the heads up

DenWav (approved):

PR #2578, opened 2026-01-30
Fix translation folding when the args are given with an explicit array

O7410:
Before:
image
After:
image

### DIFF
--- a/src/main/kotlin/translations/identification/TranslationIdentifier.kt
+++ b/src/main/kotlin/translations/identification/TranslationIdentifier.kt
@@ -31,8 +31,8 @@ import com.demonwav.mcdev.util.constantStringValue
 import com.demonwav.mcdev.util.constantValue
 import com.demonwav.mcdev.util.descriptor
 import com.demonwav.mcdev.util.findModule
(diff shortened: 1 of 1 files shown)


src/main/kotlin/translations/identification/TranslationIdentifier.kt
```diff
@@ -54,7 +54,8 @@ import org.jetbrains.uast.UMethod
 import org.jetbrains.uast.UQualifiedReferenceExpression
 import org.jetbrains.uast.evaluateString
 import org.jetbrains.uast.getContainingUClass
-import org.jetbrains.uast.util.isArrayInitializer
+import org.jetbrains.uast.kotlin.psi.toEllipsisTypeIfNeeded
```
Earthcomputer: Can you check if this is from the kotlin plugin? We can't use methods from the kotlin plugin as that creates a hard dependency

O7410:
I made some more changes

O7410:
@Earthcomputer bump

Earthcomputer (approved):

PR #2595, opened 2026-03-21
Fix potential deadlock when triggering project refresh while holding write lock

### LINKED ISSUE #2519: IntelliJ freezing upon loading multiple projects

2025.2-1.8.6

### IntelliJ version

2025.2.2 Build #IC-252.26199.169

### Operating System

Windows 10 version 22h2 build 19045.6332

### Target platform

NeoForge

### Description of the bug

As of recently when I try to open a secondary project from the recent projects section there's about a 70% chance it'll fail to load and lock up my IntelliJ forcing me to kill the process in the task manager.

All I can see when it happens is a gray (dark mode) window with the loading throbber that is stuck. Any other IntelliJ window will also have locked up making me unable to open any menu until I force IntelliJ close and restart it

### LINKED ISSUE #2566: IDE freezes when opening existing project

2025.2-1.8.11

### IntelliJ version

2025.2.5

### Operating System

Linux

### Target platform

Minecraft Forge

### Description of the bug

My IDE froze when trying to open a project. It was the first one being opened in this run of the IDE, but I have had the same issue occur even when it's not the first project. Stacktrace from `jstack` is attached. I am not sure how to reproduce this issue as it occurs extremely rarely - it is probably related to one of the other deadlock issues already opened, but not 100% sure as it happened on project open rather than when interacting with the project.

[stack.txt](https://github.com/user-attachments/files/24348252/stack.txt)

embeddedt:
The common problem among all the stacktraces I looked at is a deadlock occuring when a thread holding the write lock tries to access the AWT or the write intent lock. My reading of the IntelliJ [threading model](https://plugins.jetbrains.com/docs/intellij/threading-model.html) documentation suggests this can be problematic. Since both of the two cases I found appear to be simple (triggering a project refresh), it seems that the simplest fix is to just queue the refresh call to be run later on the correct thread, independently of any write lock held by the surrounding method.

It will take some time to confirm if this actually fixes the problem, because it is impossible to reproduce reliably, hence why I resorted to stacktrace analysis.

In theory, this fixes #2566, fixes #2519 (there may also be other issue numbers I'm not aware of)

### DIFF
--- a/src/main/kotlin/facet/MinecraftFacet.kt
+++ b/src/main/kotlin/facet/MinecraftFacet.kt
@@ -35,6 +35,7 @@ import com.intellij.facet.FacetManager
 import com.intellij.facet.FacetTypeId
 import com.intellij.facet.FacetTypeRegistry
 import com.intellij.ide.projectView.ProjectView
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.runReadAction
 import com.intellij.openapi.application.runWriteAction
 import com.intellij.openapi.module.Module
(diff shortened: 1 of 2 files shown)


Earthcomputer:
Since your other PR contained the use of LLMs (nothing wrong with that if you yourself are competent and verify yourself that it's good), could you also let me know if you used LLMs for this PR as well?

embeddedt:
I used an LLM to point me in the direction of the locks, but ended up implementing the fix independently (its suggested solution was unnecessarily complex).

DenWav:
Seems to work in my testing and the changes look sane. Thanks.

PR #2596, opened 2026-03-21
Fix mixin autocomplete spinning indefinitely in classes with many lambdas

embeddedt:
In large classes with many lambdas (e.g. Minecraft 1.20.1 `ChunkMap`/`Minecraft`/etc) mixin autocomplete for filling in `at` of `@Inject(at = "...")` stops working. More precisely, from profiling and dumping thread states, it appears to get stuck doing a ton of Psi resolution work in `hasSyntheticMethod`.

The cleanest approach I was able to find is to adjust the lambda counting code so that the indexes stored in `SourceCodeLocationInfo` can be used without requiring the expensive `hasSyntheticMethod` check in the Psi visitor. This reduces the work done during visiting to just the `Matcher`'s counting, which is not really super efficient either, but fast enough not to be an issue for interactive editing.

`hasSyntheticMethod` is now unused, but I've left it in place for the time being since it's a utility method. (Happy to remove if desired.)

Note: in the interest of full disclosure, the analysis for this problem was done in collaboration with an LLM. The fix appears straightforward and I have confirmed that it solves the problem. That said, the ASM/Psi analysis code here is quite complex and it's possible that changing the meaning of `lambdaCount` as done here may have broken something inadvertently.

### DIFF
--- a/src/main/kotlin/platform/mixin/util/AsmUtil.kt
+++ b/src/main/kotlin/platform/mixin/util/AsmUtil.kt
@@ -30,7 +30,6 @@ import com.demonwav.mcdev.util.findMethods
 import com.demonwav.mcdev.util.findModule
 import com.demonwav.mcdev.util.findQualifiedClass
 import com.demonwav.mcdev.util.fullQualifiedName
-import com.demonwav.mcdev.util.hasSyntheticMethod
 import com.demonwav.mcdev.util.innerIndexAndName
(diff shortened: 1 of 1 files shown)


DenWav:
I am not able to reproduce this issue, but the change suggested in the PR appears valid and does seem to be more efficient. I'll do some more testing to verify behavior works as intended, if so then I'll merge.

PR #2598, opened 2026-03-24
New: Support `@Shadow` constructors.

LlamaLad7:
Will be introduced in FabricMixin along with enum extensions. We don't need to bother checking whether the feature is supported on the current Mixin, because the user will already get a compile error about Shadow not being applicable to constructors.


LlamaLad7:
Given there's quite a lot of other stuff needed for enum extensions may as well hold off for now

PR #2607, opened 2026-04-11
Add support for template version ranges

Strokkur424:
Also improves the error message from "Sowwy no templates found" to a proper "This version is not supported." message.


Earthcomputer (approved):
I want to start thinking about moving this to semver in the future but this is good enough for now.

PR #2608, opened 2026-04-13
Fix valid class tweaker files initially erroring until reopened

Juuxel:
This slightly hacky fix utilises a "fake token" that matches real-looking line starts. It's unused in the parser so the user will still see an error when needed as the `INCOMPLETE_ENTRY_START` tokens are always illegal.

The exact issue seems to be with the `BAD_CHARACTER` returns for valid *characters*, since the characters themselves can be valid later when the keyword is fully typed.


Earthcomputer (approved):
Thanks a lot!

PR #2613, opened 2026-04-21
Add completion contributor for Bukkit event handlers

Leguan16:
Adds a completion contributor for Bukkit event handlers. 

It only works in a class that implements `org.bukkit.event.Listener`.

It uses templates to generate the method and provide opportunity to change method name and parameter name. 

https://github.com/user-attachments/assets/39d1b1d7-b902-4eb8-ae59-635cbb80d717

I initially created this as a poc after someone (Strokkur) posted a video of him doing something similar and wanted to try if i can implement the same logic using the plugin SDK. 

If this ends up getting added we can also add Strokkur as co-author for the idea if he wants to, i don't mind that.

### DIFF
--- a/src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
+++ b/src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
@@ -1,0 +1,80 @@
+package com.demonwav.mcdev.platform.bukkit.completion
+
+import com.intellij.codeInsight.completion.CompletionParameters
+import com.intellij.codeInsight.completion.CompletionProvider
+import com.intellij.codeInsight.completion.CompletionResultSet
+import com.intellij.codeInsight.completion.PrioritizedLookupElement
+import com.intellij.codeInsight.lookup.LookupElementBuilder
+import com.intellij.icons.AllIcons
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiClass
+import com.intellij.psi.PsiMethod
+import com.intellij.psi.PsiModifier
+import com.intellij.psi.impl.JavaPsiFacadeEx
+import com.intellij.psi.search.GlobalSearchScope
+import com.intellij.psi.search.searches.ClassInheritorsSearch
+import com.intellij.psi.util.PsiTreeUtil
+import com.intellij.util.ProcessingContext
+import org.jetbrains.annotations.NotNull
+
+class BukkitEventHandlerCompletionProvider : CompletionProvider<CompletionParameters>() {
+    companion object {
+        const val EVENT_LISTENER = "org.bukkit.event.Listener"
+        private const val BUKKIT_EVENT_FQN = "org.bukkit.event.Event"
+    }
+
+    override fun addCompletions(
+        @NotNull completionParameters: CompletionParameters,
+        @NotNull processingContext: ProcessingContext,
+        @NotNull completionResultSet: CompletionResultSet
+    ) {
+        val prefix = completionResultSet.prefixMatcher.prefix
+        if (!prefix.startsWith("on") || prefix.length == 2) return
+
+        val position = completionParameters.position
+        val containingClass = PsiTreeUtil.getParentOfType(position, PsiClass::class.java) ?: return
+        val project: Project = position.project
+        val facade = JavaPsiFacadeEx.getInstanceEx(project)
+
+        val eventListenerClass = facade.findClass(EVENT_LISTENER, GlobalSearchScope.allScope(project)) ?: return
+        if (!containingClass.isInheritor(eventListenerClass, true)) return
+        if (PsiTreeUtil.getParentOfType(position, PsiMethod::class.java) != null) return
+
+        val scope = GlobalSearchScope.allScope(project)
+        val eventBaseClass = facade.findClass(BUKKIT_EVENT_FQN, scope) ?: return
+
+        val eventNameFilter = prefix.substring(2).lowercase()
+
+        ClassInheritorsSearch.search(eventBaseClass, scope, true)
+            .forEach { psiClass ->
+                if (psiClass.isInterface || psiClass.hasModifierProperty(PsiModifier.ABSTRACT)) {
+                    return@forEach
+                }
+
+                val eventSimpleName = psiClass.name ?: return@forEach
+                if (eventNameFilter.isNotEmpty() && !eventSimpleName.lowercase().startsWith(eventNameFilter)) {
+                    return@forEach
+                }
+
+                val lookupString = "on$eventSimpleName"
+                val methodName = lookupString.replace("Event", "")
+                val qualifiedName = psiClass.qualifiedName
+
+                val element = LookupElementBuilder
+                    .create(lookupString)
+                    .withPresentableText("$lookupString()")
+                    .withTailText(" - $qualifiedName", true)
+                    .withTypeText("@EventHandler")
+                    .withIcon(AllIcons.Nodes.Method)
+                    .withBaseLookupString(lookupString)
+                    .withBoldness(true)
+                    .withInsertHandler(BukkitEventHandlerInsertHandler(methodName, qualifiedName))
+
+                completionResultSet.addElement(
+                    PrioritizedLookupElement.withPriority(element, 100.0)
+                )
+            }
+
+        completionResultSet.stopHere()
+    }
+}
--- a/src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
+++ b/src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
@@ -1,3 +1,23 @@
+/*
+ * Minecraft Development for IntelliJ
+ *
+ * https://mcdev.io/
+ *
+ * Copyright (C) 2025 minecraft-dev
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, version 3.0 only.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
 package com.demonwav.mcdev.platform.bukkit.completion
 
 import com.demonwav.mcdev.platform.bukkit.util.BukkitConstants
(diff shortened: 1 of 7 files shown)


Strokkur424:
That's cool

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+class BukkitEventHandlerCompletionProvider : CompletionProvider<CompletionParameters>() {
+    companion object {
+        const val EVENT_LISTENER = "org.bukkit.event.Listener"
+        private const val BUKKIT_EVENT_FQN = "org.bukkit.event.Event"
+    }
+
+    override fun addCompletions(
+        @NotNull completionParameters: CompletionParameters,
+        @NotNull processingContext: ProcessingContext,
+        @NotNull completionResultSet: CompletionResultSet
```
Earthcomputer: `@NotNull` is unnecessary in Kotlin, types are not null by default.

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+    ) {
+        val prefix = completionResultSet.prefixMatcher.prefix
+        if (!prefix.startsWith("on") || prefix.length == 2) return
```
Earthcomputer: Please avoid single-line if statements (including in the rest of the file)

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+
+        val position = completionParameters.position
+        val containingClass = PsiTreeUtil.getParentOfType(position, PsiClass::class.java) ?: return
```
Earthcomputer:
```suggestion
        val containingClass = position.findContainingClass() ?: return
```

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+        val project: Project = position.project
```
Earthcomputer:
```suggestion
        val project = position.project
```

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+        val facade = JavaPsiFacadeEx.getInstanceEx(project)
```
Earthcomputer: Is there any reason this needs the extended facade?

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+
+        val eventListenerClass = facade.findClass(EVENT_LISTENER, GlobalSearchScope.allScope(project)) ?: return
+        if (!containingClass.isInheritor(eventListenerClass, true)) return
+        if (PsiTreeUtil.getParentOfType(position, PsiMethod::class.java) != null) return
```
Earthcomputer: `position.findContainingMethod()`

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+
+        val scope = GlobalSearchScope.allScope(project)
+        val eventBaseClass = facade.findClass(BUKKIT_EVENT_FQN, scope) ?: return
+
+        val eventNameFilter = prefix.substring(2).lowercase()
+
+        ClassInheritorsSearch.search(eventBaseClass, scope, true)
```
Earthcomputer, line 68: This looks expensive. However it's not so easy to cache effectively because the easiest way to do that is using the PSI modification count as the cache key, and since you're in a completion provider that is changing on every keystroke. Leave it as it is for now.

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
```diff
@@ -0,0 +1,80 @@
...
+                    return@forEach
+                }
+
+                val eventSimpleName = psiClass.name ?: return@forEach
+                if (eventNameFilter.isNotEmpty() && !eventSimpleName.lowercase().startsWith(eventNameFilter)) {
+                    return@forEach
+                }
+
+                val lookupString = "on$eventSimpleName"
+                val methodName = lookupString.replace("Event", "")
```
Earthcomputer:
```suggestion
                val methodName = lookupString.removeSuffix("Event")
```

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerInsertHandler.kt
```diff
@@ -0,0 +1,45 @@
...
+class BukkitEventHandlerInsertHandler(
+    private val methodName: String,
+    private val qualifiedName: @NlsSafe String?
+) : InsertHandler<LookupElement> {
+
+    companion object {
+        private const val EVENT_HANDLER_FQN = "org.bukkit.event.EventHandler"
+    }
+
+    override fun handleInsert(@NotNull insertionContext: InsertionContext, @NotNull lookupElement: LookupElement) {
```
Earthcomputer: unnecessary `@NotNull`

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerInsertHandler.kt
```diff
@@ -0,0 +1,45 @@
...
+        template.addTextSegment("($qualifiedName ")
+        template.addVariable("EVENT_PARAM", TextExpression("event"), true)
+        template.addTextSegment(") { \n")
+        template.addEndVariable()
+        template.addTextSegment("\n}")
+
+        insertionContext.getDocument().deleteString(
+            insertionContext.getStartOffset(),
+            insertionContext.getTailOffset()
+        );
```
Earthcomputer: Unnecessary semicolon

Earthcomputer (changes requested):
In addition to my specific comments:
- Please make sure to include license headers in all files.
- The constants (inside the companion objects) should go in `BukkitConstants`. I believe some of them may already be there.

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerInsertHandler.kt
```diff
@@ -0,0 +1,45 @@
...
+    override fun handleInsert(@NotNull insertionContext: InsertionContext, @NotNull lookupElement: LookupElement) {
+        val project = insertionContext.project
+        val editor = insertionContext.editor
+
+        val templateManager = TemplateManager.getInstance(project)
+        val template = templateManager.createTemplate("", "")
+        template.isToReformat = true
+
+        template.addTextSegment("@${EVENT_HANDLER_FQN}\n")
+        template.addTextSegment("public void ")
```
Strokkur424: I see that the format of this template is hard-coded. Some people might want a different format, like have their methods be package-private (like I do). Some people might also want a different default method name (I've seen people put `on` as every handler's method name), or even different default event name. Perhaps it could be beneficial to expose properly editing the template in some way? IntelliJ already provides an interface for users to edit templates for these things, so perhaps you could hook into that and provide the method names/event params as built-in variables instead?

Strokkur424 (changes requested):

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerInsertHandler.kt
Leguan16: I will read into it thanks for making me aware of this.

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
Leguan16: Sorry for all these unnecessary elements. I initially wrote this in Java and converted it to Kotlin afterwards. I will address this.

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerInsertHandler.kt
Leguan16: I did play around a bit but I don't think i can implement this in a clean way. Especially with using existing logic. 

It definitely should be possible but I simply am not familiar enough with kotlin to do it.

Leguan16:
I think i addressed everything in the requested changes. For the editable template I am not familiar enough with Kotlin to add this and currently don't have the time for it. If someone else wants to take a look feel free to do so.

If there is anything else that needs to be changed let me know.

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerInsertHandler.kt
Strokkur424: I might play around with this and send a patch if I succeed, no guarantees though.

Leguan16:
Alright Strokkur added an editable live template so the auto completed method is now fully customizable. 

https://github.com/user-attachments/assets/d3ccca19-35ec-43ba-bfe1-6982f5120ba3

I think - from my side - this is now ready for testing and another review. Let me know if you want anything changed.

Strokkur424 (approved):

src/main/kotlin/platform/bukkit/completion/BukkitEventHandlerCompletionProvider.kt
```diff
@@ -0,0 +1,99 @@
...
+                    .withBaseLookupString(lookupString)
+                    .withBoldness(true)
+                    .withInsertHandler(BukkitEventHandlerInsertHandler(psiClass))
+
+                completionResultSet.addElement(
+                    PrioritizedLookupElement.withPriority(element, 100.0)
+                )
+            }
+
+        completionResultSet.stopHere()
```
Earthcomputer: What's the reason for this `stopHere`?

Leguan16: There is no specific reason. I added this because if i want to complete event handlers I don't need generic completions in the list. But I can remove it so they are included again if that is what is preferred.

Earthcomputer: Yes, we shouldn't be suppressing other completions without a good specific reason

Earthcomputer:
Also, can you make sure that the copyright headers for new files say 2026 rather than 2025?

Leguan16:
Oh lol yeah, I just copied them from another file and assumed they were already correct.

Leguan16:
Alright, i have updated the year and no longer suppress other completion providers

Earthcomputer (approved):

PR #2617, opened 2026-05-01
Add dynamic selectors to method completion

Bawnorton:
Previously custom selectors do not appear in the `method` autocomplete suggestion window, this adds their ids to the autocomplete suggestions (below normal methods) and optionally populates additional annotations (I have implemented this for `@Desc` based on the handling of `MIXINEXTRAS:EXPRESSION`). I would perhaps prefer to put this in the selector definition itself rather than the parser, however, that would require a bit of a redesign of how custom selectors exist in mcdev so I left it as-is.

Example of what this looks like, where before neither `@Desc` nor `@MixinSquared:Handler` would appear as a suggestion

https://github.com/user-attachments/assets/754dee3d-8f1e-4091-bb41-0cce949a18eb

### DIFF
--- a/src/main/kotlin/platform/mixin/reference/MixinSelectors.kt
+++ b/src/main/kotlin/platform/mixin/reference/MixinSelectors.kt
@@ -36,6 +36,7 @@ import com.demonwav.mcdev.util.constantStringValue
 import com.demonwav.mcdev.util.descriptor
 import com.demonwav.mcdev.util.findAnnotation
 import com.demonwav.mcdev.util.findContainingClass
+import com.demonwav.mcdev.util.findContainingModifierList
 import com.demonwav.mcdev.util.findField
 import com.demonwav.mcdev.util.findMethods
 import com.demonwav.mcdev.util.findQualifiedClass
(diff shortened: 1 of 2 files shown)


src/main/kotlin/platform/mixin/reference/MixinSelectors.kt
```diff
@@ -546,6 +555,32 @@ class DescSelectorParser : DynamicSelectorParser("Desc", "mixin:Desc") {
...
+        // add imports and reformat
+        JavaCodeStyleManager.getInstance(project).shortenClassReferences(descAnnotation)
+        JavaCodeStyleManager.getInstance(project).optimizeImports(modifierList.containingFile)
+        val formattedModifierList = CodeStyleManager.getInstance(project).reformat(modifierList) as PsiModifierList
+
+        // move the caret to @Desc("<caret>")
+        val formattedExprAnnotation = formattedModifierList.findAnnotation(DESC)
+            ?: return
+        val exprLiteral = formattedExprAnnotation.findDeclaredAttributeValue(null) ?: return
+        editor.caretModel.moveToOffset(exprLiteral.textRange.startOffset + 1)
```
Earthcomputer:
```suggestion
        val formattedDescAnnotation = formattedModifierList.findAnnotation(DESC)
            ?: return
        val descLiteral = formattedDescAnnotation.findDeclaredAttributeValue(null) ?: return
        editor.caretModel.moveToOffset(descLiteral.textRange.startOffset + 1)
```

Bawnorton: whoops, missed that, thanks

Earthcomputer (approved):

PR #2618, opened 2026-05-01
Pass owner to navigator when it can be inferred from the target

Bawnorton:
Following on from #2616 

As it turns out simply transforming the result of `getCustomOwner` was almost enough by itself, I just did it wrong in testing before 😅, however, when it comes to navigating to the target there were a couple of issues. Notably injection points would rely on the untransformed class passed from the `AtResolver` (providing the selector with enough context to transform that per #2616 would be a bit of a pain).
https://github.com/minecraft-dev/MinecraftDev/blob/66f57a651940ef2588bab9b001edd1c781e690e4/src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt#L230

 This results in the injection points looking for the target in the wrong class, so instead, where possible we trust the qualifier of the `target` (for `INVOKE`/`INVOKE_ASSIGN`/`FIELD`) to find the element, which ultimately seems to have done the trick for most injection points. 

Notably `HEAD` would wind up getting stuck in the target's injector annotation itself, resulting in a target resolution failure, so I opted to skip annotation visiting on that injection point navigator as there isn't any reason `HEAD` should point there anyway.

The injection points `CONSTANT`/`NEW`/`CTOR_HEAD` worked out the box which was convenient, however `MIXINEXTRAS:EXPRESSION` has a an issue that any definitions that reference a member "belonging" to the mixin will fail to navigate to from the gutter icon as eventually the qualifier will be queried on said member by ME's matching algorithm which will resolve as the mixin rather than the qualifier in the definition. If I were to make the same change to ME's matching it'd be to trust the definition qualifier when looking for the member rather than the AST's qualifier, however, I don't know the best way to implement that as there's a lot going on there.

This fails:
Screenshot 2026-05-01 at 21 22 30

This succeeds:
Screenshot 2026-05-01 at 21 28 10

Also fixes source file lookup when referencing same-project / sibling-project class, as sometimes the file is already decompiled when you get it via `stubClass.containingFile`

### DIFF
--- a/src/main/kotlin/platform/mixin/handlers/injectionPoint/FieldInjectionPoint.kt
+++ b/src/main/kotlin/platform/mixin/handlers/injectionPoint/FieldInjectionPoint.kt
@@ -29,6 +29,7 @@ import com.intellij.codeInsight.completion.JavaLookupElementBuilder
 import com.intellij.codeInsight.lookup.LookupElementBuilder
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
+import com.intellij.psi.JavaPsiFacade
 import com.intellij.psi.PsiAnnotation
 import com.intellij.psi.PsiArrayAccessExpression
 import com.intellij.psi.PsiClass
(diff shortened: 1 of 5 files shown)


Bawnorton:
Actually scratch this, I have a better idea. I'll open a separate pr for the class lookup fix

PR #2627, opened 2026-07-06
add MapColor preview-color gutter icons

### LINKED ISSUE #2557: Append color preview in MapColor

-

Image

Kilip1000:
Resolves #2557.

(I do not know much Kotlin and I don't know whether I could've placed some code in more apropriate locations)

### DIFF
--- a/src/main/kotlin/insight/MapColorLineMarkerProvider.kt
+++ b/src/main/kotlin/insight/MapColorLineMarkerProvider.kt
@@ -1,3 +1,23 @@
+/*
+ * Minecraft Development for IntelliJ
+ *
+ * https://mcdev.io/
+ *
+ * Copyright (C) 2026 minecraft-dev
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, version 3.0 only.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/insight/MapColorLineMarkerProvider.kt
```diff
@@ -0,0 +1,9 @@
+package com.demonwav.mcdev.insight
```
DenWav, line 1: You're missing the license headers, you need to run `./gradlew licenseFormat`

DenWav (changes requested):

src/main/kotlin/insight/MapColorLineMarkerProvider.kt
Kilip1000: It doesn't seem to change any files!
image

src/main/kotlin/insight/MapColorUtil.kt
```diff
@@ -0,0 +1,27 @@
+package com.demonwav.mcdev.insight
+
```
Earthcomputer, line 22: Still missing license header here

Earthcomputer (changes requested):

Earthcomputer (approved):
We can merge this for now. My long term plan is to have mcdev be able to infer if something is a colour from unpick data, but for that we first need to get unpick working in Fabric again.

PR #2630, opened 2026-07-16
Detect Frameworks after smart mode ready

### LINKED ISSUE #2475: Fix frequent “Detecting Minecraft frameworks” with heavy system load.

2025.1-1.8.4

When a project with more than 100 modules is open, “Detecting Minecraft frameworks” runs periodically, which puts a heavy load on the system.
I would like this to happen either faster or to be cached somehow.

ghostflyby:
A performance improvement.

Use cases of `LibraryPresentationProvider` by the IDE itself only matches the jar name by regex, proving that they are not intended for file iterations.

Framework kind are useless when the index is not ready.

All tests passed.

closes #2475


DenWav:
In concept I like this change, but it is a lot. There's several merges here, can you rebase your branch to get a clean set of commits in this PR?

DenWav:
Much nicer, thanks.

src/main/kotlin/platform/adventure/framework/AdventureLibraryDetector.kt
```diff
@@ -20,32 +20,21 @@
...
-import com.intellij.openapi.roots.libraries.LibraryPresentationProvider
-import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.openapi.project.Project
+import com.intellij.psi.search.GlobalSearchScope
 import java.util.jar.Attributes.Name.SPECIFICATION_TITLE
-import java.util.jar.Manifest
 
-class AdventurePresentationProvider : LibraryPresentationProvider<DummyLibraryProperties>(ADVENTURE_LIBRARY_KIND) {
-    override fun getIcon(properties: DummyLibraryProperties?) = PlatformAssets.ADVENTURE_ICON
+class AdventureLibraryDetector : MinecraftLibraryDetector {
```
DenWav, line 32: All of these classes were renamed to `*LibraryDetector`, but the file name is still `*PresentationProvider`. The file name should match in this case.

src/main/kotlin/facet/MinecraftLibraryDetector.kt
```diff
@@ -0,0 +1,112 @@
...
+package com.demonwav.mcdev.facet
+
+import com.demonwav.mcdev.platform.PlatformType
+import com.intellij.openapi.extensions.ExtensionPointName
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.psi.JavaPsiFacade
+import com.intellij.psi.search.FilenameIndex
+import com.intellij.psi.search.GlobalSearchScope
+import java.util.*
```
DenWav, line 30: No wildcard imports please.

DenWav (changes requested):
I'm concerned about handling of `CancellationException`, but I haven't looked deeply enough into this PR yet, will do so when I get the time.

PR #2633, opened 2026-07-25
Support mixin targets reference and completion

Enaium:

### DIFF
--- a/src/main/kotlin/platform/mixin/completion/MixinCompletionContributor.kt
+++ b/src/main/kotlin/platform/mixin/completion/MixinCompletionContributor.kt
@@ -20,8 +20,10 @@
 
 package com.demonwav.mcdev.platform.mixin.completion
 
+import com.demonwav.mcdev.platform.mixin.util.MixinConstants.Annotations.MIXIN
 import com.demonwav.mcdev.platform.mixin.util.findShadowTargets
 import com.demonwav.mcdev.platform.mixin.util.isMixin
+import com.demonwav.mcdev.util.constantStringValue
 import com.demonwav.mcdev.util.equivalentTo
 import com.demonwav.mcdev.util.filter
 import com.demonwav.mcdev.util.findContainingClass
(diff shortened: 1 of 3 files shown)


src/main/kotlin/platform/mixin/completion/MixinCompletionContributor.kt
```diff
@@ -123,4 +143,86 @@ class MixinCompletionContributor : CompletionContributor() {
...
+                        0.5,
+                    ),
+                )
+            }
+        }
+
+        // When input is empty, also show all class names (limited)
+        val cache = PsiShortNamesCache.getInstance(project)
+        if (packageName.isEmpty() && text.firstOrNull()?.isUpperCase() == true) {
+            for (className in cache.allClassNames) {
```
DenWav, line 211: I'm not sure this is a good idea. I'm concerned this could be too slow on slower machines. I think the easiest option would be to use `AllClassesGetter.processJavaClasses`, which would effectively do the same thing, but with process cancellation checks handled for you.

After testing the PR I do think this is way too slow to be useable. `AllClassesGetter.processJavaClasses` performs much better.

src/main/kotlin/platform/mixin/completion/MixinCompletionContributor.kt
```diff
@@ -59,6 +73,12 @@ class MixinCompletionContributor : CompletionContributor() {
             return
         }
 
+        // Check if completing inside @Mixin targets attribute
+        if (isInsideMixinTargets(position)) {
```
DenWav, line 77: Unless I'm missing something this call (and the whole delegating method) can be use the `MixinTargetsReference.ELEMENT_PATTERN`. You just need to get the `PsiLiteral` that position points to: `val literal = PsiTreeUtil.getParentOfType(position, PsiLiteral::class.java)`.

Then just `MixinTargetsReference.ELEMENT_PATTERN.accepts(literal)`.

DenWav (changes requested):
So the biggest problem with this PR is it implements a lot of stuff from scratch where it isn't necessary. I threw a few comments in with some of the stuff I noticed, but then I realized that the reference contributor could just extend `JavaClassReferenceProvider`.

Then I kind of rabbit-holed looking at this and implemented a couple of other things to make it work nicer. One important bit is raising the priority of `MixinCompletionContributor`, with it on `last` priority it wasn't actually running.

I didn't want to steal your thunder or anything, so I pushed my stuff up to a separate branch: https://github.com/minecraft-dev/MinecraftDev/tree/mixin-target-ref

Take a look if you like.

Enaium:
No need to argue; if my project's pr needs any changes, I will assist in making those changes.

Enaium:
Could you add me as a co-author of the commit 23c36a5be58b526e5ae6fe6591cd78de5bceb6cf?

DenWav:
Yes, I will. Thank you for your contribution. :)

DenWav:
Merged in https://github.com/minecraft-dev/MinecraftDev/commit/c0908cb848852942c919dcdfd356433032ba3f8b, with co-author credit.

PR #2639, opened 2026-08-16
Add Fabric 26.1+ class templates (Mojmap)

Orlisan:
Added new FreeMarker templates (.ft) for Fabric 26.1+, using Mojmap instead of Yarn, since recent Minecraft versions ship deobfuscated and Yarn mappings are no longer used. Registered them in MinecraftTemplates.kt.

Added FabricModule.computeVersion(), which reads the Minecraft version from depends.minecraft in fabric.mod.json (handles both the single-string and array forms allowed by the Fabric spec), since minecraftVersion from McpModuleType only applies to Forge. MinecraftClassCreateAction uses this to pick the correct Fabric templates based on the module's actual Minecraft version.

Tested locally on a Fabric 26.2 mod project — new templates are generated correctly.


TrygveK:
The mappings is good, but I think your AI might have hallucinated a bit on the enchantment stuff because enchantments are entirely data-driven (JSON) in 1.21+, so extending the Java class doesn't actually work anymore

In file FabricEnchantment (26.1+).java.ft here:
```public class ${NAME} extends Enchantment {
    public ${NAME}(Rarity weight, EnchantmentTarget type, EquipmentSlot[] slotTypes) {
        super(weight, type, slotTypes);
    }
}
```

Orlisan:
@TrygveK i know, i commented that doesn't work, but i don't remove that because its a feature made by others and i can't delete direcly without asking, (and i not used AI, i copied the neoforge template)

TrygveK:
> @TrygveK i know, i commented that doesn't work, but i don't remove that because its a feature made by others and i can't delete direcly without asking, (and i not used AI, i copied the neoforge template)

Makes zero sense, last time I checked when GitHub says an + icon beside the file it meant **YOU** made it
image
You said you commented that it didn't work, but your initial PR description claims: "Tested locally on a Fabric 26.2 mod project — new templates are generated correctly."
And you aren't deleting someone else's work by removing it. Also, copying the NeoForge template doesn't change the fact that Java enchantments are dead in 1.21+. Adding templates that generate broken code just looks like an AI hallucinated the wrong version
Beside what template did you even use?

Orlisan:
My apologies, you're right — that part of the PR description was an AI hallucination. I was aware the enchantment template doesn't work under 1.21+'s data-driven system; I left it as I found it and only translated it to Mojmap, without actually fixing the underlying issue.

What I meant about "someone else's work" was different from how it sounded: even though I wrote that file, it's a Mojmap translation of an existing template, not something I designed from scratch — I wasn't trying to say I copied it from someone else's PR, and that file is the old file renamed

Either way, I should have flagged this clearly instead of leaving it silently broken. Sorry for the confusion.

(This response is ai translated in english, may contains allucinations)

TrygveK:
Translation tools don't usually hallucinate but anyway, since you agree that the file is broken for 1.21+, please just fix that Java template from this. And since you didn't probably actually test it then a bunch of things is probably broken

Orlisan:
ok, tomorrow i'll fix

TrygveK:
I just ran this branch locally in the sandbox to see what it generates, and it doesn't generate anything it completely crashes the IDE plugin when you click New > Minecraft Class .

The FabricModule.computeVersion() method you added throws a NumberFormatException because it's trying to parse an empty string as an integer when checking the version:
```
Java
java.lang.NumberFormatException: For input string: ""
    at com.demonwav.mcdev.util.SemanticVersion$Companion.parse(SemanticVersion.kt:191)
    at com.demonwav.mcdev.platform.fabric.FabricModule.computeVersion(FabricModule.kt:92)
```
There is no way this was 'tested locally' like the PR description claimed. The version logic is fundamentally broken, and the Java enchantment template is dead for 1.21+

Orlisan:
weird, i tested it and it works, can you describe exactly what you do to meet the exception?

TrygveK:
Run ./gradlew runIde to start the sandbox.

Create a new Fabric project targeting Minecraft 26.2.

Wait for the project's Gradle sync to finish completely.

Right-click the main Java package (src/main/java/...).

Click New > Minecraft Class.

The IDE throws the NumberFormatException and the plugin crashes

Orlisan:
what's the depends -> minecraft value in fabric.mod.json

TrygveK:
It says "minecraft": "${minecraft_version}".

Your computeVersion method is trying to parse the raw Gradle placeholder string instead of an actual version number. When it fails to find a number, it passes "" to the integer parser and crashes so the plugin needs to gracefully handle unresolved variables

Orlisan:
ok understood, tomorrow i'll fix

TrygveK:
after hardcoding the loader version and minecraft ver it generates 
```package com.minecraft.www;
//Not functional for MC 1.20.3+

import net.minecraft.world.entity.EquipmentSlot;
import net.minecraft.world.item.enchantment.Enchantment;
import net.minecraft.world.item.enchantment.EnchantmentCategory;

public class test extends Enchantment {
    public test(Rarity rarity, EnchantmentCategory category, EquipmentSlot[] slots) {
        super(rarity, category, slots);
    }
}
```
```
Expected 4 arguments but found 3
Cannot resolve symbol 'EnchantmentCategory'
Cannot resolve symbol 'Rarity'
Cannot inherit from record 'net.minecraft.world.item.enchantment.Enchantment'
Cannot resolve symbol 'EnchantmentCategory'
```
and your AI even said "//Not functional for MC 1.20.3+" so like why you trying to put this into an modern Fabric 26.2

TrygveK (changes requested):
Tested this branch locally in the sandbox. Before this can be merged, several critical issues need to be addressed:

1. Version Parsing Crash (NumberFormatException):
When generating on a standard project using Gradle variables ("minecraft": "${minecraft_version}"), the version parser fails to handle the placeholder and throws a fatal exception, completely crashing the IDE plugin:
```
java.lang.NumberFormatException: For input string: ""
    at com.demonwav.mcdev.util.SemanticVersion$Companion.parse(SemanticVersion.kt:191)
    at com.demonwav.mcdev.platform.fabric.FabricModule.computeVersion(FabricModule.kt:92)
```
The version logic needs to safely handle unresolved template variables instead of crashing.

2. Obsolete Template Logic:
Once the version is hardcoded to bypass the crash, the generated Java code explicitly confesses that it is broken: //Not functional for MC 1.20.3+.

3. Modern Compatibility Errors:
Attempting to use this template on a modern Fabric project (like 26.2) fails immediately with compilation errors because enchantments are now data-driven JSON files rather than legacy Java classes:
```
Cannot resolve symbol 'EnchantmentCategory'
Cannot resolve symbol 'Rarity'
Cannot inherit from record 'net.minecraft.world.item.enchantment.Enchantment'
Cannot resolve symbol 'EnchantmentCategory'
Expected 4 arguments but found 3
```
Please fix the version parser to prevent the IDE crash, and update/replace the enchantment template so it actually aligns with modern 1.21+ data-driven systems instead of generating dead code.

Orlisan:
I think that i fix that, but i'm not sure because my hardware's limits cannot test me very well

TrygveK (changes requested):
While the ```NumberFormatException``` crash is fixed, simply removing the Enchantment option for 1.21+ doesn't actually solve the core issue this PR was trying to address. Modern 1.21+ enchantments are now data-driven JSON files rather than Java classes and instead of disabling the feature, can you update the template to generate the correct JSON data structure for 1.21+ enchantments so the feature is actually supported in modern versions?

src/main/resources/META-INF/plugin.xml
```diff
@@ -1751,6 +1751,10 @@
                 text="Minecraft Class" description="Create skeleton classes used in Minecraft Mods">
```
TrygveK: This one isn't necessary anymore and make bad user experience as you have one that does both, please don't just copy things from chatgpt. Understand it

Orlisan, line 1751: I didn't copy that from ChatGPT, and the line you commented on wasn't something I modified in this PR — so I'm not sure what you'd like me to change there. Could you clarify which part you mean, why isn't necessary and what i supposed to do?
Also, just to explain the design: I kept Class creation and Json creation as two separate actions because isAvailable runs before the user picks a kind (Block/Item/Enchantment/etc.), so I can't conditionally show the Json action only when they're about to pick Enchantment. And merging them into one action would mean mixing Java source-root and resource-root logic under the same isAvailable, which felt wrong since they're fundamentally different targets. Happy to reconsider if you see a cleaner way to do this.

src/main/resources/META-INF/plugin.xml
```diff
@@ -1751,6 +1751,10 @@
...
             <add-to-group group-id="NewGroup" anchor="before" relative-to-action="NewFile"/>
         </action>
+        <action class="com.demonwav.mcdev.insight.generation.MinecraftResourceCreateAction" id="CreateMinecraftJson"
+                text="Minecraft Class" description="Create skeleton jsons used in Minecraft Mods and datapacks">
```
TrygveK:
```suggestion
           
```
It's going to be confusing having it twice

Orlisan:
ehm so? I can't see anymore the last changes comment

TrygveK (changes requested):

src/main/resources/META-INF/plugin.xml
Orlisan: and what i have to do instead? they are different action inheritors from different classes, if you consider that a new class is wasted for a single use i already made an expansion to her in my other branch for another PR

Earthcomputer:
This is for the old project creator, which will probably be removed soon anyway. The new one is what most people use, and I believe that's been fixed for 26.1+ already.

Orlisan:
this isn't for any project creator @Earthcomputer

Earthcomputer:
Oh I see sorry. That's what I get for glancing

src/main/resources/META-INF/plugin.xml
Orlisan: if the problem was the <text field, i done

Earthcomputer:
What's the state of this PR? Is it done? The commented-out code would need removing in that case

TrygveK:
Idk lost track of that after everything with the AI it's probably not done

> What's the state of this PR? Is it done? The commented-out code would need removing in that case

Orlisan:
for the last time, i don't used ai, and i think the mantainer work includes finding exactly what don't work and telling me what i have to fix, not tell a generic "probably not done"

TrygveK:
> for the last time, i don't used ai, and i think the mantainer work includes finding exactly what don't work and telling me what i have to fix, not tell a generic "probably not done"

a maintainer's role is to review code to ensure it meets project standards. It is the contributor's responsibility to submit clean, functional, and tested code not something broken that an maintainer has to fix themselves

Orlisan:
As I mentioned, I don't have the hardware to test every scenario this could affect. I've cleaned up the leftover commented code. What specifically do you need from me to get this mergeable

Earthcomputer:
Well that's what I was asking you

Earthcomputer:
If you say it's done and works properly just say that lol

Orlisan:
i think that works, but i thought it a week ago too, but at the time for example i didn't know fabric can use gradle placeholders, another ignorance or mistake can be now, and since i can't test every scenario, i'm not, and i'll never be sure
The things I was told to fix have been fixed

Orlisan:
so?

Orlisan:
From what I understood, I was rightly called out for things like the crash on gradle placeholders, a copy-paste error (or at least that's what I assumed the issue was, since the request wasn't very clear and I didn't get an explanation), and keeping a broken feature while admitting it was broken. All three issues have been fixed:

- Added support for gradle placeholders ${...}
- Added an action for json resources in assets — a feature that, since it's a bit underused right now, I plan to expand in another PR I've actually already made, now named New -> Minecraft Enchantment (I was then thinking of turning it into Minecraft Json)
- Changed the text field in plugin.xml for that action from Minecraft Class to Minecraft Json

There may be other issues worth fixing; if so, I'd appreciate it if they were pointed out, and otherwise I'd be glad if the PR could be merged.

Orlisan:
Anyway, this PR is a bugfix. If you never merge it, the plugin will remain broken for the newest minecraft versions of Fabric mods.

TrygveK (commented):
Removing the commented text doesn't resolve the bug.

By deleting the mcVersion check entirely, now executes unconditionally for every user. This means the 1.21 JSON action will incorrectly appear in 1.18, 1.20, and all other legacy projects.

This action must be version-gated. You need to actively implement a working mcVersion >= 1.21 check here so it doesn't break for older projects

src/main/kotlin/insight/generation/MinecraftResourceCreateAction.kt
TrygveK, line 1: This

Earthcomputer: There is a check in there to be fair

Orlisan:
I tested with a Fabric 1.18 project generated via the plugin wizard: the "New Minecraft Enchantment" action correctly does not appear, confirming the >= 1.21 check in isAvailable works as expected. The full condition requires the directory to be under a resource root, the modid to be resolvable from the path, AND the version to be non-null and >= 1.21 — by short-circuit evaluation, a failed/null version detection alone is enough to safely return false. The commented-out code I removed was, per the diff, only dead comments — I never touched or removed the mcVersion check itself.
Anyway, what would this bug be — I don't see any mention of unresolved bugs in this chat."

Earthcomputer (changes requested):
You need to start co-operating with the reviewers, or else I will have to close the PR and let someone else pick it up. We've told you need to debug it yourself but you refused to do it.

Maybe it is true you didn't use AI, but either way this PR spends a lot of effort reimplementing something that we already have implemented better (Minecraft version detection), which is something that AI tends to do, hence the prior assumption. Regardless of the use of AI, this problem needs to be fixed as well as other irrelevant code changes.

Orlisan:
i'm trying to collaborate with revievewrs, and my minecraft version detection change was done because original mcVersion in current MinecraftClassCreateAction resulted me null, so i supposed, it's was only for forge, also because that was only use of that variable.
Now i tested for older and newest versions

Earthcomputer:
It shouldn't be null, if it is you should try to work out why. On Fabric it's assigned [here](https://github.com/minecraft-dev/MinecraftDev/blob/156fe70647ee7103872cad1279bb194d6ab77e10/src/main/kotlin/platform/mcp/fabricloom/FabricLoomProjectResolverExtension.kt#L59) so you should try to work out why that isn't being assigned properly for you.

Orlisan:
Now it works without manual detection. Probably my machine was just slow to finish the Gradle sync, so the version wasn't populated in time for my earlier test.

Earthcomputer (changes requested):
This is looking much better now, thank you for improving it. Can you remove the now unused imports that appeared as a result of removing some code?

And, can you remove the json enchantment template? It can always be added later if people want it, but ideally you generate enchantments using datagen rather than writing json files manually.

Orlisan:
For the unused imports, I'll do it as soon as I can. As for removing the JSON enchantment template, I get the concern about a dedicated button for a niche feature, that's fair, and it's actually part of why I was planning a follow-up PR to expand this action with more templates and JSON structures, so it wouldn't stay a single-purpose button. That said, I want to be upfront that this specific part was added at TrygveK's explicit request, to replace the old pre-1.21 class-based generation I had removed, and I worked on it specifically for that purpose. Having it reverted now, from a different reviewer with the opposite request, is a bit frustrating.

Earthcomputer:
I get that it's frustrating, but that's just how it is sometimes. TrygveK is not a maintainer, they are just someone who decided to review the PR, and to be fair they did overall a good job. Reviews from maintainers should usually take precedence.

Orlisan:
so in short I have to throw away a day of work and my entire other branch?

Earthcomputer:
If something is going to take you an entire day to implement, it's good to check with the maintainers if it's something that is actually wanted, to avoid wasting time on something that isn't wanted. I'm sorry that you wasted time, but I can't be held responsible for someone else's review.

Orlisan:
ok, i'll remove it, can you tell me if in a separated pr that, with block, items, json structures and singular files like blockstates, and similar things can be conceptually accepted?

Earthcomputer:
These files should really be created with [data generation](https://docs.fabricmc.net/develop/data-generation/setup), even for relatively small mods, so I'm inclined to say no.

Orlisan:
okok, also considering that data generation is not instantaneous to setup and that for small/medium sized mods it is not worth it?

Earthcomputer:
It's part of the mod generator on the Fabric site

Orlisan:
ok, i give up

src/main/resources/fileTemplates/j2ee/skeleton/forge/ForgeEnchantment (1.17+).java.ft
```diff
@@ -4,9 +4,8 @@
 import net.minecraft.world.entity.EquipmentSlot;
 import net.minecraft.world.item.enchantment.Enchantment;
 import net.minecraft.world.item.enchantment.EnchantmentCategory;
-
 public class ${NAME} extends Enchantment {
```
Earthcomputer, line 8: Please don't delete this blank line

src/main/resources/messages/MinecraftDevelopment.properties
```diff
@@ -180,6 +180,7 @@ generate.event_listener.ignore_if_canceled=Ignore if event is canceled
 generate.class.caption=Minecraft Class
 generate.class.description=Class generation for modders
 
+
```
Earthcomputer: ?

src/main/kotlin/util/SemanticVersion.kt
Earthcomputer, line 1: Please revert this file

src/main/kotlin/platform/AbstractModule.kt
Earthcomputer, line 1: Please revert this file

src/main/kotlin/platform/fabric/FabricModule.kt
Earthcomputer, line 1: Please revert this file

src/main/kotlin/insight/generation/MinecraftClassCreateAction.kt
```diff
@@ -140,7 +151,7 @@ class MinecraftClassCreateAction :
         return JavaDirectoryService.getInstance().createClass(dir, className, templateName, false)
     }
 
-    private class ClassInputValidator(
+    class ClassInputValidator(
```
Earthcomputer: Is there a reason anymore for this change?

src/main/resources/fileTemplates/j2ee/skeleton/neoforge/NeoForgeEnchantment.java.ft
Earthcomputer, line 1: Please revert this file

Earthcomputer (changes requested):
Still some unrelated formatting changes

src/main/kotlin/insight/generation/MinecraftClassCreateAction.kt
Orlisan: No, I probably didn't realize I had changed it.

Earthcomputer (approved):

Orlisan:
Thanks for the approval and for your patience! Is anything else needed before the merge takes place, or is it just a matter of timing?

Earthcomputer:
I'm going to test it and then merge it

Orlisan:
ok

PR #2642, opened 2026-08-28
Fix: Support debugging interpolated expression strings.

LlamaLad7:
image

### DIFF
--- a/src/main/kotlin/platform/mixin/expression/gui/MEShowFlowAction.kt
+++ b/src/main/kotlin/platform/mixin/expression/gui/MEShowFlowAction.kt
@@ -20,11 +20,11 @@
 
 package com.demonwav.mcdev.platform.mixin.expression.gui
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/platform/mixin/expression/gui/MEShowFlowAction.kt
```diff
@@ -158,6 +156,22 @@ class MEShowFlowAction : AnAction() {
         }
     }
 
+    private fun findExpressionString(anchor: PsiElement): PsiElement? {
+        val nameValue = anchor.parentOfType<PsiNameValuePair>() ?: return null
+        if (anchor === nameValue.nameIdentifier) {
```
Earthcomputer: Wouldn't this allow the `=` token to be the anchor? I think it makes more sense to check if the value is an ancestor of the anchor.

Earthcomputer (approved):

PR #2643, opened 2026-09-02
Feature/mixin quantifiers

### LINKED ISSUE #2518: Mixin wildcard * is not properly accounted for

2025.1-1.8.6

### IntelliJ version

2025.1.5.1

### Operating System

macOS

### Target platform

Mixins

### Description of the bug

When targeting multiple methods using one handler with `method = "*"`, MCDev suddenly produces unnecessary red error lines (the code works fine and the game runs correctly).
To reproduce:
1. Take any mixin handler (This method of reproduction works with most handlers I've tested; however, some handlers work properly and do not produce the red lines. I find that WrapOperation and ModifyExpressionValue are the most consistent when reproducing.)
2. Replace the full method qualifier with "*" instead
3. Your mixin should now look something similar to
```java
@ModifyExpressionValue(method = "*", at = @At(value = "CONSTANT", args = "intValue=2048"))
    private int useHeightInstead(int original, @Local(argsOnly = true)BlockEntity blockEntity, @Local BeamEmitter.BeamSegment beamSegment) {
        return blockEntity instanceof LaserBlockEntity ? beamSegment.getHeight() : original;
    }
```
Original method: https://github.com/ekulxam/laseredstone/blob/28a1abf532e7ce0c5e47d5e75c9375ad63297e4d/src/main/java/survivalblock/laseredstone/mixin/BeaconBlockEntityRendererMixin.java#L15

Image

These red error lines can be ignored as they actually do not signify actual errors (the game runs with no crashes).

LlamaLad7:
Basic support for Mixin quantifiers, along with related bugfixes. Commits best consumed invididually.

Closes #2518

### DIFF
--- a/src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt
+++ b/src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt
@@ -36,9 +36,11 @@ import com.demonwav.mcdev.platform.mixin.util.findSourceClass
 import com.demonwav.mcdev.platform.mixin.util.findSourceElement
 import com.demonwav.mcdev.platform.mixin.util.isClinit
 import com.demonwav.mcdev.platform.mixin.util.memberReference
+import com.demonwav.mcdev.util.Quantifier
 import com.demonwav.mcdev.util.computeStringArray
 import com.demonwav.mcdev.util.constantStringValue
 import com.demonwav.mcdev.util.constantValue
+import com.demonwav.mcdev.util.countIsAtLeast
 import com.demonwav.mcdev.util.descriptor
 import com.demonwav.mcdev.util.equivalentTo
 import com.demonwav.mcdev.util.findMethods
@@ -197,7 +199,17 @@ class AtResolver(
                 InsnResolutionInfo.Failure()
             }
         }
-        return collectVisitor.visit(targetMethod) as? InsnResolutionInfo.Failure
+        return when (val result = collectVisitor.visit(targetMethod)) {
+            is InsnResolutionInfo.Failure -> result
+            is InsnResolutionInfo.Success -> {
+                val minMatches = collectVisitor.quantifier.min(Quantifier.Context.INSTRUCTION).coerceAtLeast(1)
+                if (result.results.countIsAtLeast(minMatches)) {
+                    null
+                } else {
+                    InsnResolutionInfo.Failure()
+                }
+            }
+        }
     }
 
     fun resolveInstructions(
--- a/src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt
+++ b/src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt
@@ -176,6 +176,8 @@ class AtResolver(
                 else -> 0
             }
         }
+
+        const val DEFAULT_UNRESOLVED_MESSAGE = "Cannot resolve any instructions in target"
     }
 
     fun isUnresolved(): InsnResolutionInfo.Failure? {
@@ -192,11 +194,11 @@ class AtResolver(
         )
         if (collectVisitor == null) {
             // syntax error in target
-            val stringValue = targetAttr?.constantStringValue ?: return InsnResolutionInfo.Failure()
+            val stringValue = targetAttr?.constantStringValue ?: return InsnResolutionInfo.Failure(DEFAULT_UNRESOLVED_MESSAGE)
             return if (isMiscDynamicSelector(at.project, stringValue)) {
                 null
             } else {
-                InsnResolutionInfo.Failure()
+                InsnResolutionInfo.Failure(DEFAULT_UNRESOLVED_MESSAGE)
             }
         }
         return when (val result = collectVisitor.visit(targetMethod)) {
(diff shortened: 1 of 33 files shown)


src/main/kotlin/util/MemberReference.kt
```diff
@@ -87,90 +79,39 @@ data class MemberReference(
 
     override fun matchField(owner: String, name: String, desc: String): Boolean {
         assert(!owner.contains('.'))
-        return (this.matchAllNames || this.name == name) &&
+        return canEverMatch(name) &&
```
Earthcomputer: Even though the implementation is the same, I'm not sure this is the right function to use here logically. Maybe however we can rename the function to `matchesName` or something

src/main/kotlin/platform/mixin/util/MemberInfo.kt
```diff
@@ -80,14 +60,14 @@ data class MemberInfo(
 
     override fun matchField(owner: String, name: String, desc: String): Boolean {
         assert(!owner.contains('.'))
-        return (this.matchAllNames || this.name == name) &&
+        return canEverMatch(name) &&
```
Earthcomputer: Same here.

src/main/kotlin/util/sequences.kt
```diff
@@ -50,3 +50,21 @@ fun <T> Sequence<T>.memoized(): Sequence<T> {
         }
     }
 }
+
+val Sequence<*>.countIs get() = SequenceCountCompareOps(this)
+
+class SequenceCountCompareOps(private val sequence: Sequence<*>) {
```
Earthcomputer: Not a big fan of magic syntax like this

Earthcomputer (commented):
I still need to test it, but reviewed the code

src/main/kotlin/util/sequences.kt
LlamaLad7: What would you suggest instead? It's a concept from Scala so it's at least somewhat established in that sense, but even aside from being nice to read I don't think there's any other way to expose the same functionality without 3 (or even 6) separate functions which feels overkill.

Earthcomputer: I wasn't familiar with the pattern and since it's not a Scala codebase you can't necessarily expect the average reader to be familiar with the pattern. I was also confused at the usage site and had to read the implementation to understand what it was.

I reckon that unless this is commonly needed, reverting to the previous code `.drop(1).any()` is fine

LlamaLad7: It's used in 2 other places too as `.countIs >= minMatches` or similar

LlamaLad7: It reads quite well imo

Earthcomputer: I read it as "count `i`s" initially and I was like "what is `i`?". I'm just not expecting to read code in this way.

I think implementing them as separate functions is clearer. You can express all the inequalities as `drop(?).any()` so it's at least not too many lines

LlamaLad7: Perhaps `sizeIs` would be clearer? Or would you rather `.countIsAtLeast`, `.countIsMoreThan`, etc?

Earthcomputer: Yeah let's do `countIsAtLeast`

src/main/kotlin/platform/mixin/handlers/injectionPoint/AtResolver.kt
```diff
@@ -197,7 +199,17 @@ class AtResolver(
...
         }
-        return collectVisitor.visit(targetMethod) as? InsnResolutionInfo.Failure
+        return when (val result = collectVisitor.visit(targetMethod)) {
+            is InsnResolutionInfo.Failure -> result
+            is InsnResolutionInfo.Success -> {
+                val minMatches = collectVisitor.quantifier.min(Quantifier.Context.INSTRUCTION).coerceAtLeast(1)
+                if (result.results.countIsAtLeast(minMatches)) {
+                    null
+                } else {
+                    InsnResolutionInfo.Failure()
```
Earthcomputer: Maybe we want more failure information here, so that we can give the user a better error message. The max quantifier generates a better error message by being added to `filterStats` via `CollectVisitor.visit`.

Earthcomputer (commented):
Testing this out now. First thing I noticed is that the "fix method signature" quick fix picks the first invalid method target and fixes to that, even if there is no way to satisfy all targeted methods. Ideally if there is no way to satisfy all targeted methods there should be no quick fix. The correct solution is to compute an intersection of the expected method signatures in `InvalidInjectorMethodSignatureInspection` and recommend that for the fix. But if that's too annoying, then an easy solution for now would be disabling the quick fix if there are multiple different expected signatures.

LlamaLad7:
> First thing I noticed is that the "fix method signature" quick fix picks the first invalid method target and fixes to that, even if there is no way to satisfy all targeted methods

It does, though this is existing behaviour and not caused by the PR. I agree it should be fixed, though I won't be able to make any changes until next week since I'm away. Feel free to do any final bits yourself if you want.

Earthcomputer:
> It does, though this is existing behaviour and not caused by the PR.

I'll fix it separately, it's probably important to fix now that `*` isn't so broken in mcdev anyway, especially now that people are more likely to encounter it with the upcoming lambda stuff

PR #2645, opened 2026-09-19
Fix: Handle quantifiers and staticness properly when resolving Mixin target methods.

LlamaLad7:
We should only take as many matches as is allowed, and additionally static methods should not be considered unless our handler is also static or we are trying to select exactly 1 method.


LlamaLad7:
Test failure seems to be a flake

src/main/kotlin/platform/mixin/handlers/InjectorAnnotationHandler.kt
```diff
@@ -63,21 +66,19 @@ abstract class InjectorAnnotationHandler : MixinAnnotationHandler {
         val selectors = method.mapNotNull { parseMixinSelector(it, methodAttr!!) } +
             desc.mapNotNull { DescSelectorParser.Util.descSelectorFromAnnotation(it) }
 
-        val targetClassMethods = selectors.associateWith { selector ->
-            val actualTarget = selector.getCustomOwner(targetClass)
-            (actualTarget to actualTarget.methods)
+        val targetsBySelector = selectors.associateWith { selector ->
+            selector.getCustomOwner(targetClass)
         }
+        val isStatic = annotation.findParentOfType<PsiMethod>()?.hasModifierProperty(PsiModifier.STATIC) ?: true
```
Earthcomputer: I feel like this variable should be called `allowStatic` rather than `isStatic`, as it's true if the annotation doesn't have any parent method (common while still typing the mixin).

src/main/kotlin/platform/mixin/reference/AbstractMethodReference.kt
```diff
@@ -82,13 +84,15 @@ abstract class AbstractMethodReference : PolyReferenceResolver(), MixinReference
             return false
         }
 
+        val isStatic = context.parentOfType<PsiMethod>()?.hasModifierProperty(PsiModifier.STATIC) ?: true
```
Earthcomputer: Same here

src/main/kotlin/platform/mixin/reference/AbstractMethodReference.kt
```diff
@@ -104,10 +108,13 @@ abstract class AbstractMethodReference : PolyReferenceResolver(), MixinReference
...
 
     private fun isAmbiguous(targets: Collection<ClassNode>, targetReference: MemberInfo): Boolean {
-        return targets.any { it.findMethods(targetReference.withQuantifier(Quantifier.Any)).countIsAtLeast(2) }
+        return targets.any {
+            it.findMethods(targetReference.withQuantifier(Quantifier.Any), allowStatic = true).countIsAtLeast(2)
+        }
     }
 
     fun resolve(context: PsiElement): Sequence<ClassAndMethodNode>? {
+        val isStatic = context.parentOfType<PsiMethod>()?.hasModifierProperty(PsiModifier.STATIC) ?: true
```
Earthcomputer: Same here

Earthcomputer (commented):
Should probably also check `handler.canAlwaysBeStatic`

LlamaLad7:
> Should probably also check `handler.canAlwaysBeStatic`

Not sure what you mean by this

Earthcomputer:
There's a method in `InjectorAnnotationHandler` called `canAlwaysBeStatic`, it's true most of the time but can be false on stock mixin. When it's false, the handler method can't be static when the target method is also non-static

LlamaLad7:
I'm aware, but Mixin does not account for that when selecting target methods. The incorrect staticness will be flagged later by the relevant inspection (matching the error that will occur at runtime)

Earthcomputer (approved):

